PR #18, opened 2016-07-19
Aml 23 information and guidance

dashton82:
Change made so that it goes to the first step to inform the user from invoking the action to create an account.

Also included a change so that the worker role will work on azure, adding in the connection information and the tokenizer values for substitution at release


dashton82:
Sat with @floydprice

PR #54, opened 2016-08-03
[AML-237] First/last name transposed

ijrussell:
[AML-237] First/last name transposed


dashton82:
@floydprice ive looked but needed to fix something else, can you be the third set of eyes? also we should be using SingleOrDefault instead of FirstOrDefault

ijrussell:
DO NOT MERGE

PR #56, opened 2016-08-04
AML-243 - Queue name renaming for environment consistency

dashton82:
Update all queue names to remove the das_at_eas_ part from the queue name. Also removed an non-used class that was being used to define the queue names before the MessagePolicy was introduced.


OutKa5t:
[TestResult.txt](https://github.com/SkillsFundingAgency/das-employerapprenticeshipsservice/files/401998/TestResult.txt)

PR #83, opened 2016-08-12
Data service queue work

robinoconnor:
Have merge the latest with master. When I run the RunBuild  script its failing but I think the errors are due to local environment being setup incorrect for Levy.

The update is quite small and so I don't think the issue are with the code changes. Might be worth running the RunBuild script on your machine when checking this code to make sure its just my setup and not a code issue.

The code adds support for a message to be sent to the queue the Data project is listening on so that when an account is created the Data project's worker can then request the account information via the Levy API.


robinoconnor:
Sorted out the merge conflicts with master

PR #100, opened 2016-08-17
Update content for all.cshtml

PeteKowalczyk:


PeteKowalczyk:
Changed "Role" to "What you can done" -- but this requires me to change the roles themselves before this page makes sense.

PR #109, opened 2016-08-30
AML-278 - Access denied error

dashton82:
Update to the base controller to try and get the accountid from the request so that the user can be redirected back to the correct page.


robinoconnor:
All looks good from what I can see code wise.

PR #110, opened 2016-08-31
Fixed the bug by moving where we replace the empRef value and now it …

robinoconnor:
…should be replaced if a value can be found for the test user email. Have also fixed a naming bug with the levy database


robinoconnor:
There is also a fix in one of the tests as well that had to be added due tot he logic change in the code itself. Now if the test data file has a match it will replace the HMRC EmpRef code even if it has been fully populated by the HMRC service.

PR #113, opened 2016-09-08
Commitments view list ui

leewadhams:


mgwilliam:
the api client class looks like it has some useful base functions we'll need in other clients - consider extracting a base class

mgwilliam:
the commands should be using domain entities not client entities (the orchestrator should be mapping)

mgwilliam:
how do we ensure the account id (that's posted as a hidden form field when creating a commitment) hasn't been spoofed?

mgwilliam:
CreateCommitmentHandler doesn't follow naming convention ("*CommandHandler") and also looks like a query rather than a command

PR #123, opened 2016-09-12
Submit commitment

leewadhams:
Submission of a commitment and creation of task triggered.
Note: There is a todo outstanding around configuration. Same issue as the Commitments Api config.


src/SFA.DAS.EmployerApprenticeshipsService.Web/DependencyResolution/DefaultRegistry.cs
```diff
@@ -60,6 +61,9 @@ public class DefaultRegistry : Registry {
             }
 
             RegisterMediator();
+
+            // TODO: LWA - Is there another way?
+            For<ITasksApi>().Use<TasksApi>().Ctor<string>().Is("http://localhost:21482/");
```
ijrussell: Yes there is! We'll add this to the backlog.

src/SFA.DAS.Tasks.Api.Client/TasksApi.cs
```diff
@@ -0,0 +1,52 @@
...
+using Newtonsoft.Json;
+using Task = SFA.DAS.Tasks.Domain.Entities.Task;
+
+namespace SFA.DAS.Tasks.Api.Client
+{
+    public class TasksApi : HttpClientBase, ITasksApi
+    {
+        private readonly string _baseUrl;
+
+        public TasksApi(string baseUrl)
```
ijrussell: This will change as part of the changes for DI

PR #140, opened 2016-09-20
Additional prod configuration

OutKa5t:
Ive removed RDP so its clean for now. Pushed another cscfg and modified the build to accept it with VSTS workflow (for now)


OutKa5t:
Self organisation rules

PR #167, opened 2016-09-26
Ui tidy up

leewadhams:
Tidy up of UI from walk-through.
Added a extension method for GDS dates. Will apply other dates in solution if agree it's a decent idea.


src/SFA.DAS.EmployerApprenticeshipsService.Web/Extensions/DateTimeExtensions.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace System
+{
+    public static class DateTimeExtensions
```
ijrussell: +1 -> I've added this to Provider as well.

ijrussell:
Fix the conflict(s) and it's good to go.

PR #225, opened 2016-10-18
Aml 557 homepage validation

jamesmking:
- Added a Render Section on the layout to allow me to define page/view specific javascript
- Changed the validation to the radio buttons on the homepage - users can only create an account if they have selected 'yes' or shown the error message


dashton82:
@jamesmking  can we put the script in the bundle? give us a shout if you need a hand

jamesmking:
@dashton82 Thanks Dan, created a bundle and pushed an update

PR #241, opened 2016-10-26
Commitments validation

leewadhams:
Validation on creation of commitment and addition of apprenticeship. Legal Entities now using radio buttons instead of drop down.


src/SFA.DAS.EmployerApprenticeshipsService.Web/Content/elements.css
```diff
@@ -2597,10 +2597,13 @@ input[type=number] {
 
 .validation-summary,.validation-summary-errors {
     background: #fef7f7;
-    border-left: 3px solid #df3034;
+<<<<<<< HEAD
```
ijrussell: Remove merge residue

PR #246, opened 2016-10-27
AML 384 english percentage

dashton82:
Rounding is now done as per story. This may need to be reviewed to always round down


src/SFA.DAS.LevyAggregationProvider.Worker.UnitTests/Providers/LevyAggregationTests/WhenIAggregateLevyData.cs
```diff
@@ -223,8 +224,8 @@ public void ThenTheTotalIncludesTheTopUpValueAndTheEnglishPercentage()
             var actual = _levyAggregator.BuildAggregate(actualData);
 
             //Assert
-            var expectedTotal = (actualData.Data[0].LevyDueYtd * actualData.Data[0].EnglishFraction) + (actualData.Data[0].TopUp * actualData.Data[0].EnglishFraction);
-            Assert.AreEqual(expectedTotal, actual.Data[0].Balance);
+            var expectedTotal = Math.Round(actualData.Data[0].LevyDueYtd * actualData.Data[0].EnglishFraction,2) + Math.Round(actualData.Data[0].TopUp * actualData.Data[0].EnglishFraction,2);
```
robinoconnor: Could you not just have round the final value of expectedTotal rather than round the two values and then have to round them again afterwards?

dashton82: the round early is due to the previous amount already being rounded. This is the correct place for it, however - we are going to delete all this later 👍

robinoconnor: ok cool, thought I'd check. I'll merge it now

PR #290, opened 2016-11-08
Validate ui

leewadhams:
- Removed client side validation
- Fixed validation bug on blank provider search


leewadhams:
Replaced by another pull request

PR #292, opened 2016-11-08
refactoring Framework Id to be a string

scottcowan:
There is a collision issue with integer framework Ids and the new format for DAS is 403-8-10 so it needs to be a string for this to work.

I've removed the int.Parses and changed ITrainingProgramme


src/SFA.DAS.EAS.Web/Views/EmployerCommitments/EditApprenticeshipEntry.cshtml
```diff
@@ -1,116 +1,101 @@
 ﻿@using SFA.DAS.EAS.Web.Extensions
 @model SFA.DAS.EAS.Web.Models.ApprenticeshipViewModel
 
-@*@section pageSpecificJS {
```
scottcowan: this looks like a bad merge, I'll clear it up

PR #295, opened 2016-11-09
New way of deploying WebJobs to handle more than one

EwanNoble:
Schedules are now managed through the settings.job file in the WebJob project root using a CRON expression.
WebJobs are built into the Host.WebJob project and deployed from there.

Successfully demo'd in AT and Test environments


dashton82:
Looks good to me @OutKa5t do you want to cast your eyes over before i merge it in?

PR #308, opened 2016-11-15
Aml 685 payment worker

dashton82:
New worker role for the processing of payments


src/SFA.DAS.EmployerApprenticeshipsService.Application/Commands/Payments/RefreshPaymentData/RefreshPaymentDataCommandHandler.cs
```diff
@@ -0,0 +1,58 @@
...
+            {
+                throw new InvalidRequestException(validationResult.ValidationDictionary);
+            }
+
+            var payments = await _paymentsEventsApiClient.GetPayments(message.PeriodEnd, message.AccountId.ToString());
+            var sendPaymentDataChanged = false;
+
+            foreach (var payment in payments.Items)
+            {
+                var existingPayment = await _dasLevyRepository.GetPaymentData(Guid.Parse(payment.Id));
```
robinoconnor: Do we need to get payment data individually or can we just get all the current payment IDs and then take out all payments that are currently in the DB? Seem it would be less load on the DB and would only require a single DB call

robinoconnor: Agree, for now we just need to add this as a tech task given it doesn't just affect this story. I'll add a tech task now to review this sort of issue system wide.

PR #313, opened 2016-11-16
Approving commitments

froberg:


mgwilliam:
can you add some todos in the code to remove those magic strings like "save-no-send"

PR #319, opened 2016-11-21
Fe 2b2

froberg:


src/SFA.DAS.EmployerApprenticeshipsService.Application/Commands/SubmitCommitment/SubmitCommitmentCommandHandler.cs
```diff
@@ -39,7 +39,7 @@ public SubmitCommitmentCommandHandler(ICommitmentsApi commitmentApi, ITasksApi t
                 throw new InvalidRequestException(new Dictionary<string, string> { { "Commitment", "This commiment does not belong to this Employer Account " } });
 
             // TODO: Refactor out these magic strings
-            var agreementStatus = (message.SaveOrSend == "send-approve") 
+            var agreementStatus = (message.SaveOrSend.Contains("approve"))
```
leewadhams: When it's a "Provider will add details" journey we don't have any apprenticeships to change the status of. Do we still need to call it? Does any other status need to change?

PR #325, opened 2016-11-22
Improvements on training list

froberg:
- Added script for drop down
- Sort list
- format title


froberg:
@rajeshdigital can you have a  look at the UI for this and same for provider PR (Dropdown trainings).
We can have a look at it if you have time tomorrow?

// Chris

PR #333, opened 2016-11-24
Bug new fields not saving

froberg:


src/SFA.DAS.EAS.Web/Views/EmployerCommitments/Details.cshtml
```diff
@@ -59,15 +61,24 @@
...
+                            @if (apprenticeship.StartDate != null && apprenticeship.EndDate != null)
+                            {
+                                <span>
+                                    @apprenticeship.StartDate.Value.ToGdsFormatWithoutDay() to @apprenticeship.EndDate.Value.ToGdsFormatWithoutDay()
+                                </span>
+                            }
+                        </td>
+                        <td>
+                            @($"{apprenticeship.Cost:#}") &nbsp;
+                        </td>
```
mgwilliam: format as monetary value

PR #339, opened 2016-11-25
Employer ui changes from 21 nov 2016

rajeshdigital:


src/SFA.DAS.EAS.Web/Views/EmployerCommitments/Details.cshtml
```diff
@@ -6,28 +6,36 @@
...
 
+<div class="grid-row">
+    <div class="column-full">
         <h2 class="heading-medium">@GetApprenticesCountText(Model.Apprenticeships)</h2>
         
         @{
             var finishEditingText = Model.ShowApproveOnlyOption ? "Continue to approval" : "Finish editing";
         }
-
+        <p><hr /></p>
```
jamesmking: <hr> tags can exist on their own, don't need to be wrapped in a <p>

src/SFA.DAS.EAS.Web/Views/EmployerCommitments/Acknowledgement.cshtml
```diff
@@ -4,7 +4,7 @@
     <div class="column-two-thirds">
 
 
-        <div class="govuk-box-highlight">
+        <div class="govuk-box-highlight green-box">
```
jamesmking: In the rest of the AML journey, all error summaries and green flash messages sit in a column-full so are 100% width of the page

PR #344, opened 2016-11-25
Aml 749 logging

dashton82:
Updated logging in EAS to use the Redis NLog target. @kalim81 has been informed of the new app names that need to be added to the elasticsearch index. 

@EwanNoble if you could check the config that would be great


EwanNoble:
Config looks fine to me

PR #353, opened 2016-11-28
updated events API

froberg:


mgwilliam:
looks like merge is in progress? see unit test project file and packages file

froberg:
@mgwilliam Ok, that is fixed now

PR #360, opened 2016-12-02
Account id update

robinoconnor:
Can you both review these changes from and AML and Commitments prespective and then once you're both happy we can merge it into master.

Thanks


leewadhams:
This looks good from a Commitments point of view.

PR #404, opened 2016-12-15
Dropdown accessibility fix

rajeshdigital:


src/SFA.DAS.EAS.Web/Web.config
```diff
@@ -29,7 +29,7 @@
     <add key="RedisConnection" value="localhost" />
   </appSettings>
   <system.web>
-    <compilation debug="true" targetFramework="4.5" />
+    <compilation debug="false" targetFramework="4.5" />
```
jamesmking: I don't think we want this on master?

PR #431, opened 2017-01-06
Aml 814 change account name

chrisfoster76:


src/SFA.DAS.EmployerApprenticeshipsService.Application/Commands/RenameEmployerAccount/RenameEmployerAccountCommandHandler.cs
```diff
@@ -0,0 +1,40 @@
...
+        public RenameEmployerAccountCommandHandler(IEmployerAccountRepository accountRepository, IValidator<RenameEmployerAccountCommand> validator, IHashingService hashingService)
+        {
+            _accountRepository = accountRepository;
+            _validator = validator;
+            _hashingService = hashingService;
+        }
+
+        protected override async Task HandleCore(RenameEmployerAccountCommand message)
+        {
+            var validationResult = _validator.Validate(message);
```
dashton82: Might be worth in the future changing this to ValidateAsync and validating as to wether the user is an Owner

PR #435, opened 2017-01-10
Add missing Cost validation

leewadhams:
Add check for Cost field length. Note: There is a Provider part to this bug too.

[DCM-164 Cost validation](https://skillsfundingagency.atlassian.net/browse/DCM-164)


src/SFA.DAS.EAS.Web/Validators/ApprenticeshipViewModelValidator.cs
```diff
@@ -12,7 +13,8 @@ public class ApprenticeshipViewModelValidator : AbstractValidator<Apprenticeship
         public ApprenticeshipViewModelValidator()
         {
             var yesterday = DateTime.Now.AddDays(-1);
-            Func<string, int, bool> lengthLessThan = (str, lenth) => (str?.Length ?? 0) <= lenth;
+            Func<string, int, bool> lengthLessThan = (str, length) => (str?.Length ?? 0) <= length;
+            Func<string, int, bool> numberOfDigitsLessThan = (str, length) => { return (str?.Count(char.IsDigit) ?? 0) < length; };
```
mgwilliam: my English teach would say this should be "fewerThan" :)

PR #440, opened 2017-01-10
DCM-138 Invalid state commitment

froberg:
- Setting nocache on controller actions
- Throwing InvalidStateException if no rights to view
- Catch exception and retrun redirect to index with flash message.


froberg:
- Waiting for copy (heading and message)
- Using current flash message but the flash message (temp data) might change soon.

froberg:
https://skillsfundingagency.atlassian.net/browse/DCM-182

PR #448, opened 2017-01-12
Added account events

MichaelYoung1981:
Publish events to the events api when an account is created, renamed or a legal entity is added.

### DIFF
--- a/src/SFA.DAS.EAS.Web/SFA.DAS.EAS.Web.csproj
+++ b/src/SFA.DAS.EAS.Web/SFA.DAS.EAS.Web.csproj
@@ -18,7 +18,7 @@
     <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
     <MvcBuildViews>false</MvcBuildViews>
     <UseIISExpress>true</UseIISExpress>
-    <IISExpressSSLPort>44360</IISExpressSSLPort>
+    <IISExpressSSLPort />
     <IISExpressAnonymousAuthentication />
     <IISExpressWindowsAuthentication />
     <IISExpressUseClassicPipelineMode />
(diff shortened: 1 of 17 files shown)


src/SFA.DAS.EmployerApprenticeshipsService.Application/Commands/CreateAccount/CreateAccountCommandHandler.cs
```diff
@@ -66,6 +68,8 @@ public CreateAccountCommandHandler(IAccountRepository accountRepository, IUserRe
             {
                 AccountId = accountId
             });
+
+            await _mediator.SendAsync(new CreateAccountEventCommand { HashedAccountId = hashedAccountId, Event = "AccountCreated" });
```
dashton82: Should this be
`await _mediator.PublishAsync(new CreateAccountEvent());`
Where CreateAccountEvent should be IAsyncNotification

src/SFA.DAS.EmployerApprenticeshipsService.Application/Commands/CreateAccountEvent/CreateAccountEventCommand.cs
```diff
@@ -0,0 +1,11 @@
+﻿using MediatR;
+using SFA.DAS.EAS.Application.Commands.CreateAccount;
+
+namespace SFA.DAS.EAS.Application.Commands.CreateAccountEvent
+{
+    public class CreateAccountEventCommand : IAsyncRequest
```
dashton82: IAsyncNotification?

src/SFA.DAS.EAS.Web/SFA.DAS.EAS.Web.csproj
```diff
@@ -18,7 +18,7 @@
     <TargetFrameworkVersion>v4.5</TargetFrameworkVersion>
     <MvcBuildViews>false</MvcBuildViews>
     <UseIISExpress>true</UseIISExpress>
-    <IISExpressSSLPort />
+    <IISExpressSSLPort>44360</IISExpressSSLPort>
```
dashton82: Can this be taken out - it shouldn't use a specific port

src/SFA.DAS.EmployerApprenticeshipsService.Application/Commands/CreateAccount/CreateAccountCommandHandler.cs
floydprice: Its possible that we are guilty of doing this elsewhere but IAsynNotification should be the approach...

PR #459, opened 2017-01-13
Table fonts size   dcm 169

rajeshdigital:
Font size for table elements changed to 16px.


leewadhams:
@rajeshdigital is this going to have any effect on any other parts of the employer site? Is @jamesmking aware if it does?

jamesmking:
@leewadhams i've reviewed it, it shouldn't effect anything else on the employer site. I'm happy for it to be merged

PR #480, opened 2017-01-18
Add legal entity id to account api

MichaelYoung1981:
Added legal entity id to the data returns from the accounts information API.


dashton82:
@MichaelYoung1981 you need to merge Master into this - there is a bit of a conflict due to the database changing.

PR #494, opened 2017-01-24
Removes hard-coded "active" status for non-Companies House entities

chrisfoster76:
Captures correct status for Companies House entities
Removes some unused TempData entries
Removes entity status required field validation


src/SFA.DAS.EAS.Web/Controllers/EmployerAccountOrganisationController.cs
```diff
@@ -291,7 +257,7 @@ private void CreateOrganisationCookieData(OrchestratorResponse<OrganisationDetai
                     OrganisationName = response.Data.Name,
                     OrganisationDateOfInception = response.Data.DateOfInception,
                     OrganisationRegisteredAddress = response.Data.Address,
-                    OrganisationStatus = response.Data.Status ?? "active",
+                    OrganisationStatus = response.Data.Status ?? "",
```
robinoconnor: Change to string.empty

PR #502, opened 2017-01-26
Changed accounts api to be more restul.

MichaelYoung1981:


src/SFA.DAS.EAS.Account.Api.Types/LegalEntityViewModel.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+    public class LegalEntityViewModel : IAccountResource
+    {
+        public long Id { get; set; }
+        public string RegisteredAddress { get; set; }
+        public string Source { get; set; }
+        public string CompanyStatus { get; set; }
+        public string Name { get; set; }
+        public string Code { get; set; }
+        public DateTime? DateOfIncorporation { get; set; }
```
robinoconnor: We have changed the 'DateOfIncorporation' to 'DateOfInception' now as legal entities are now not just companies. Also 'CompanyStatus' should now just be 'Status' and 'RegisteredAddress' can now just be 'Address'

benilovj: @robinoconnor regarding `DateOfInception`, are we actually getting this data for anything other than companies in CH?

robinoconnor: At the moment we only get this from companies house however in future we cannot rule out that we may get these date for other types of organisations later on. Given not organisations are companies it would be misleading to return a generic legal entity with a non generic term given this is an API and may be consumed by others later on.

src/SFA.DAS.EAS.Account.Api.Types/ResourceViewModel.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace SFA.DAS.EAS.Account.Api.Types
+{
+    public class ResourceViewModel : IAccountResource
+    {
+        public string Id { get; set; }
+        public string Href { get; set; }
```
robinoconnor: Change Href to Uri (worth doing this for all occurrences as this is the correct term used with non HTML code for .net

src/SFA.DAS.EAS.Api/Orchestrators/AccountsOrchestrator.cs
```diff
@@ -44,11 +43,88 @@ public AccountsOrchestrator(IMediator mediator, ILogger logger )
...
+            };
+
+            return payeSchemeViewModel;
+        }
+
+        private LegalEntityViewModel ConvertLegalEntityToViewModel(GetLegalEntityByIdResponse legalEntityResult)
+        {
+            var legalEntityViewModel = new LegalEntityViewModel
+            {
+                DateOfIncorporation = legalEntityResult.LegalEntity.DateOfIncorporation,
```
robinoconnor: Just a quick note, we use automapper in the projects so you can use this for your mapping to save on extra code. We are aware some people don't like to use it which is fine however its there and being used if you wish to use it. It will save you quite a bit of time for simple mapping

src/SFA.DAS.EmployerApprenticeshipsService.Application/Queries/GetEmployerAccountByHashedId/GetEmployerAccountByHashedIdValidator.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+    public class GetEmployerAccountByHashedIdValidator : IValidator<GetEmployerAccountByHashedIdQuery>
+    {
+        public ValidationResult Validate(GetEmployerAccountByHashedIdQuery item)
+        {
+            var validationResult = new ValidationResult();
+
+            if (string.IsNullOrEmpty(item.HashedAccountId))
+            {
+                validationResult.AddError(nameof(item.HashedAccountId));
```
robinoconnor: Worth adding to the validation message that the error is that the hashedAccountId is empty or null as there may be additional validation added later on and it will help debug without looking into the code.

PR #505, opened 2017-01-30
Model namespace refactoring

robinoconnor:


src/SFA.DAS.EAS.Web/Controllers/VisualTestController.cs
```diff
@@ -266,10 +269,10 @@ public VisualTestController()
...
-                {"~/Views/Home/FakeUserSignIn.cshtml", new SignInUserViewModel ()
+                {"~/Views/Home/FakeUserSignIn.cshtml", new ViewModels.SignInUserViewModel ()
                     {
-                        AvailableUsers = new List<SignInUserModel>() {
-                            new SignInUserModel() {Email = "a@b.com", FirstName = "Bojack", LastName = "Horseman", UserId="123", UserSelected= "foo" }
+                        AvailableUsers = new List<UserViewModel>() {
+                            new UserViewModel() {Email = "a@b.com", FirstName = "Bojack", LastName = "Horseman", UserId="123", UserSelected= "foo" }
                         }
                     }},
                 {"~/Views/Home/Index.cshtml", new OrchestratorResponse<UserAccountsViewModel>()
```
dashton82: We should delete this controller

PR #525, opened 2017-02-02
DCM-119-b Redis Session storage config and nuget

froberg:
Do we need to set up something for the configuration transform?
/c


floydprice:
What are you guys using session state for? We have resisted making the application "State-full" until now... would be interested in the use case ;-)

mgwilliam:
@floydprice it's to support use of tempdata for flash messages that need a redirect.

leewadhams:
Closing this request as we're no longer going to use a shared session store. We'll be relying on sticky sessions.

PR #534, opened 2017-02-07
Dcm 175 employer commitment delete apprentice before approval

chrisfoster76:


src/SFA.DAS.EmployerApprenticeshipsService.Infrastructure/Services/CommitmentsService.cs
```diff
@@ -0,0 +1,25 @@
...
+using SFA.DAS.Commitments.Api.Client;
+using SFA.DAS.EAS.Domain.Interfaces;
+
+namespace SFA.DAS.EAS.Infrastructure.Services
+{
+    public class CommitmentsService : ICommitmentsService
+    {
+        private readonly ICommitmentsApi _commitmentsApi;
+
+        public CommitmentsService(ICommitmentsApi commitmentsApi)
```
leewadhams: We should probably have a quick chat about this. I can see why you've introduced it. We should update the other commands to use the service too if that's how we want to go.

chrisfoster76: Ok cool - does seem slightly redundant but seems to be the way they've gone elsewhere in the solution

PR #564, opened 2017-02-16
AML-1021 Add most recent agreement status to legal entity model

simonbu11:


src/SFA.DAS.EAS.Account.Api.Types/EmployerAgreementStatus.cs
```diff
@@ -0,0 +1,16 @@
...
+    public enum EmployerAgreementStatus
+    {
+        [Description("Not signed")]
+        Pending = 1,
+        [Description("Signed")]
+        Signed = 2,
+        [Description("Expired")]
+        Expired = 3,
+        [Description("Superceded")]
+        Superceded = 4
```
mgwilliam: supersede

simonbu11: Teach me to copy and paste

mgwilliam: yep, only works if copying from stack overflow

PR #577, opened 2017-02-22
Aml 958 ogd service

dashton82:


dashton82:
Can you let @EwanNoble know when this is merged so we can make sure that the correct config and APIs are in place 😃

PR #580, opened 2017-02-22
Handle the session expired flow for the Create Account journey.

floydprice:
When a user sits on the create account summary screen and her session expires before she hits confirm a login flow is triggered with a redirect of /account/create however the action was intended to be a POST but EmployerUsers does a redirect which is a GET yielding a page not found to the user... this redirects back to the Summary screen where the user can pick up the journey again.


floydprice:
If we had actual RESTful urls this would not be an issue as the url would be the same just the VERB would differ.... 👜

PR #632, opened 2017-03-17
Aml 1031 signed agreement event

robinoconnor:


src/SFA.DAS.EAS.Api/Controllers/EmployerAgreementController.cs
```diff
@@ -0,0 +1,27 @@
...
+    {
+        private readonly AgreementOrchestrator _orchestrator;
+
+        public EmployerAgreementController(AgreementOrchestrator orchestrator)
+        {
+            _orchestrator = orchestrator;
+        }
+
+        [Route("{agreementId}", Name = "AgreementById")]
+        [Authorize(Roles = "ReadAllEmployerAgreements")]
```
dashton82: Need to make sure DevOps are aware of this new role

robinoconnor: ok will do

PR #640, opened 2017-03-21
Aml 1031 signed agreement event

robinoconnor:


src/SFA.DAS.EmployerApprenticeshipsService.Application.UnitTests/Commands/SignEmployerAgreementTests/WhenISignAnEmployerAgreement.cs
```diff
@@ -33,13 +43,42 @@ public void Setup()
...
+            _hashingService.Setup(x => x.HashValue(It.IsAny<long>())).Returns(HashedLegalEntityId);
 
             _validator = new Mock<IValidator<SignEmployerAgreementCommand>>();
             _validator.Setup(x => x.ValidateAsync(It.IsAny<SignEmployerAgreementCommand>())).ReturnsAsync(new ValidationResult { ValidationDictionary = new Dictionary<string, string> ()});
 
+
+            _agreement = new EmployerAgreementView
+            {
+                HashedAgreementId = "124GHJG",
+                LegalEntityId = 56465
```
dashton82: you could put that id into the hashing service hash value for completeness, fine to leave as is

PR #642, opened 2017-03-21
Use empref from command instead of hmrc response

dashton82:
Change made so that the empref we use to store information in the
database is taken from the command, rather than the response from HMRC.
This is due to formatting problems with this information when it is
returned from the API


floydprice:
Happy for this to be merged (excellent work!) once we have spoken with HMRC...

PR #644, opened 2017-03-22
Update token service client

dashton82:


OutKa5t:
Might have an issue with the cert location but theres only one way to find out....

PR #660, opened 2017-03-29
Aml 1112 invitation audit

robinoconnor:


robinoconnor:
Had to change the audit to use the user ID and account ID and membership table doesn't have its own ID to use

PR #744, opened 2017-05-03
Aml 1175 tempdata cookie

dashton82:


dashton82:
This fixes AML-1012 as well

PR #801, opened 2017-05-24
Makes total cost field maxlength 7

chrisfoster76:


leewadhams:
So we're limiting the textbox to 7 characters for html 5 browsers. Are we happy for non-html 5 browsers to enter data > 7 characters in length. What will the consequence be?

PR #805, opened 2017-05-26
AML 190 remove org

dashton82:


src/SFA.DAS.EAS.Employer_Account.Database/StoredProcedures/GetEmployerAgreement_ByAccountIdAndLegalEntityId.sql
```diff
@@ -0,0 +1,29 @@
+﻿CREATE PROCEDURE [employer_account].[GetEmployerAgreement_ByAccountIdAndLegalEntityId]
```
MichaelYoung1981: Is this stored proc used?  I can see a repo method (GetLatestAccountLegalEntityAgreement) which calls it but that method has no usages.

src/SFA.DAS.EAS.Web/Orchestrators/EmployerAgreementOrchestrator.cs
```diff
@@ -297,5 +301,141 @@ public EmployerAgreementOrchestrator(IMediator mediator, ILogger logger, Employe
...
+                if (model.RemoveOrganisation == null)
+                {
+                    response.Status = HttpStatusCode.BadRequest;
+                    response.FlashMessage = new FlashMessageViewModel
+                    {
+                        Headline = "Errors to fix",
+                        Message = "Check the following details:",
+                        ErrorMessages = new Dictionary<string, string> { { "RemoveOrganisation", "Confirm you wish to remove the organisation" } },
+                        Severity = FlashMessageSeverityLevel.Error
+                    };
```
MichaelYoung1981: Could the FlashMessage creation be extracted out into a function to save all the duplication of strings and condense it down to a 1 line function call with a single parameter?

src/SFA.DAS.EmployerApprenticeshipsService.Application/Commands/RemoveLegalEntity/RemoveLegalEntityCommandValidator.cs
```diff
@@ -0,0 +1,88 @@
...
+using System.Threading.Tasks;
+using SFA.DAS.EAS.Application.Validation;
+using SFA.DAS.EAS.Domain.Data.Repositories;
+using SFA.DAS.EAS.Domain.Interfaces;
+using SFA.DAS.EAS.Domain.Models.EmployerAgreement;
+using SFA.DAS.EAS.Domain.Models.UserProfile;
+
+namespace SFA.DAS.EAS.Application.Commands.RemoveLegalEntity
+{
+    public class RemoveLegalEntityCommandValidator : IValidator<RemoveLegalEntityCommand>
```
MichaelYoung1981: I know this goes against how every validator in DAS has been  written but...

Could/should this class not be internal?  It'll never be instantiated outside of the application layer.  As you've tested through the command handler anyway (which is the right way to do it, IMO) it would be a straight forward change.  Obviously, it'd be inconsistent so feel free to ignroe. ;)

src/SFA.DAS.EAS.Employer_Account.Database/StoredProcedures/GetEmployerAgreement_ByAccountIdAndLegalEntityId.sql
dashton82: This will be removed - this was when i was going to do a different approach

src/SFA.DAS.EAS.Web/Orchestrators/EmployerAgreementOrchestrator.cs
dashton82: Made new method off the flash message view model for Creating and Error Flash Message view model

src/SFA.DAS.EmployerApprenticeshipsService.Application/Commands/RemoveLegalEntity/RemoveLegalEntityCommandValidator.cs
dashton82: Maybe going forward we can do this it can be internal but we'd obviously have to change the instantiating of the object.

MichaelYoung1981: Fair enough, we can take a look at it when we start the refactoring.

robinoconnor:
Need to review how we handle audit exceptions but this can be covered at a later date as we currently don't handle them anywhere.

PR #810, opened 2017-05-31
Dcm 570 payment priority

leewadhams:


froberg:
Fine, 
We can merge it tomorrow.
/c

PR #859, opened 2017-06-20
Aml 1250 org search

dashton82:


src/SFA.DAS.EmployerApprenticeshipsService.Application/Queries/GetOrganisations/GetOrganisationsQueryHandler.cs
```diff
@@ -0,0 +1,39 @@
...
+            var valdiationResult = _validator.Validate(message);
+
+            if (!valdiationResult.IsValid())
+            {
+                throw new InvalidRequestException(valdiationResult.ValidationDictionary);
+            }
+
+            var organisations = await _referenceDataService.SearchOrganisations(message.SearchTerm);
+
+            if (organisations == null)
```
robinoconnor: Does this work if the service returns an empty collection?

PR #898, opened 2017-07-04
using string format to £{}

froberg:


froberg:
Merge to  **origin/release_10-07-2017_DL-V2**  as well.

PR #911, opened 2017-07-05
.NET 4.6.2 migration

EwanNoble:


EwanNoble:
@dashton82 what effect will these Accounts API client changes have on consumers?

PR #928, opened 2017-07-12
Return education org as public body.

MichaelYoung1981:


src/SFA.DAS.EmployerApprenticeshipsService.Infrastructure/Services/ReferenceDataService.cs
```diff
@@ -144,7 +144,7 @@ private OrganisationType ConvertToOrganisationType(ReferenceData.Api.Client.Dto.
                 case ReferenceData.Api.Client.Dto.OrganisationType.Company:
                     return OrganisationType.CompaniesHouse;
                 case ReferenceData.Api.Client.Dto.OrganisationType.EducationOrganisation:
-                    return OrganisationType.Other;
+                    return OrganisationType.PublicBodies;
```
robinoconnor: Are we not having an educational type?

PR #935, opened 2017-07-17
Aml 1311 commitments move

dashton82:


dashton82:
Not to be merged until all code is up-to-date in the new solution

PR #983, opened 2017-09-04
[BUG] returning the correct employer agreement

froberg:
Several employer agreements can have the same LegalEntityId and therefor there is a risk of picking up the wrong agreement.  
I added account-id condition to make sure we only get agreements for the account calling the API.

On AT:
```
SELECT TOP 15 *
FROM employer_account.EmployerAgreement ea
WHERE ea.LegalEntityId = 18434
ORDER BY Id DESC
```
/c

### DIFF
--- a/src/SFA.DAS.EAS.Api/Orchestrators/AccountsOrchestrator.cs
+++ b/src/SFA.DAS.EAS.Api/Orchestrators/AccountsOrchestrator.cs
@@ -91,8 +91,8 @@ namespace SFA.DAS.EAS.Api.Orchestrators
         public async Task<OrchestratorResponse<LegalEntityViewModel>> GetLegalEntity(string hashedAccountId, long legalEntityId)
         {
             _logger.Info($"Getting legal entity {legalEntityId} for account {hashedAccountId}");
-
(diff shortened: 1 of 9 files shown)


src/SFA.DAS.EAS.Api/Orchestrators/AccountsOrchestrator.cs
```diff
@@ -91,8 +91,8 @@ public AccountsOrchestrator(IMediator mediator, ILog logger, IMapper mapper, IHa
         public async Task<OrchestratorResponse<LegalEntityViewModel>> GetLegalEntity(string hashedAccountId, long legalEntityId)
         {
             _logger.Info($"Getting legal entity {legalEntityId} for account {hashedAccountId}");
-
-            var legalEntityResult = await _mediator.SendAsync(new GetLegalEntityByIdQuery { Id = legalEntityId });
+            var accountId = _hashingService.DecodeValue(hashedAccountId);
+            var legalEntityResult = await _mediator.SendAsync(new GetLegalEntityByIdQuery { AccountId = accountId, Id = legalEntityId });
```
MichaelYoung1981: Could we pass the hashedAccountId to the Query and decode the value in there?  That'd be more consistent with the other queries.

PR #1001, opened 2017-09-29
Aml 1649 purple header

jamesmking:


IainStevenson:
Closing on request from testers. Will merge later

PR #1005, opened 2017-10-09
Aml 1607 wrong english fraction

TC1066:
The fix I've made for this includes adding an EnglishTraction column to the TransactionLine table. The following script is intended to fill that column in with the EnglishFraction that should have been used when the line amounts were calculated. Some will be left empty. This is a on-off script - after this the changes to Levy processing will fill in the value when the amounts are calculated.
Here's the script:
UPDATE
tl
SET
tl.EnglishFraction = td.EnglishFraction
FROM
employer_financial.TransactionLine AS tl 
INNER JOIN employer_financial.GetLevyDeclarationAndTopUp AS td 
ON tl.EmpRef = td.EmpRef 
AND tl.AccountId = td.AccountId
AND tl.SubmissionId = td.SubmissionId


EwanNoble:
Robin says merge

PR #1008, opened 2017-10-13
Aml 1610 remove commitments

MichaelYoung1981:


leewadhams:
We'll need to remove the bundles for commitments but need to wait for "sharing solution" to be deployed as commitments is currently getting all it's assets from MA.

PR #1010, opened 2017-10-16
Aml 1239 sign agreement tasks

robinoconnor:
Check the code and then we'll go through it and discuss the changes. Its a major piece of work so worth taking the time to go through it


src/SFA.DAS.EAS.Web/Controllers/EmployerTeamController.cs
```diff
@@ -45,6 +45,13 @@ public class EmployerTeamController : BaseController
...
 
         [HttpGet]
+        [Route("activity")]
+        public ActionResult Activity()
+        {
+            return View();
+        }
+
+        [HttpGet]
         [Route("view")]
```
TC1066: Why have the route different to the controller action name?

src/SFA.DAS.EAS.Web/Controllers/EmployerTeamController.cs
```diff
@@ -45,6 +45,13 @@ public class EmployerTeamController : BaseController
         }
 
         [HttpGet]
+        [Route("activity")]
+        public ActionResult Activity()
```
TC1066: Do you need the route as the action name is the same?

PR #1036, opened 2017-11-02
Implementing Web Policy to strip out sensative response headers

IainStevenson:
Implementing Web Policy to strip out sensative response headers


IainStevenson:
Somehow the titloe 'Implementation of the common OrganisationType enum' has contaiminated this after a revert was issued. the title should be something like  'Prevent_ASP.NET_Header_Information_Disclosure'

PR #1037, opened 2017-11-02
Added filter to make sure only active legal entities are returned by …

robinoconnor:
…the API


MichaelYoung1981:
Does the accounts API also use this stored proc?

PR #1042, opened 2017-11-13
Aml 1240 activities

TC1066:


src/SFA.DAS.EmployerAccounts.Events/Messages/Message.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
+
+namespace SFA.DAS.EmployerAccounts.Events.Messages
+{
+    public abstract class Message
```
robinoconnor: I think the message base class should be place in the DAS shared project message library as its something that could be commonly used by all. Only the created date field would be in it however and it should be renamed to MessageBase as its a base class.

src/SFA.DAS.EmployerAccounts.Events/Messages/Message.cs
```diff
@@ -0,0 +1,20 @@
...
+    {
+        protected Message(string signedByName=null, long accountId=0)
+        {
+            SignedByName = signedByName;
```
robinoconnor: SignedByName : this looks like its the external user id that created the message so it should be renamed to reflect this. I'd take it out of the generic message base as well and create a new message for MA that can be used when the message is created by a user as not all maybe.

src/SFA.DAS.EmployerAccounts.Events/Messages/Message.cs
```diff
@@ -0,0 +1,20 @@
...
+            AccountId = accountId;
```
robinoconnor: Same with Account ID. Move this to a base message that is used by MA as messages from other project may not use account id

src/SFA.DAS.EmployerAccounts.Events/Messages/Message.cs
```diff
@@ -0,0 +1,20 @@
...
+            PostedDatedTime = DateTime.Now;
```
robinoconnor: Rather than posted i'd say created as it may be posted at a later date.

src/SFA.DAS.EmployerAccounts.Events/Messages/AgreementSignedMessage.cs
```diff
@@ -1,12 +1,26 @@
...
+        }
+
+        public AgreementSignedMessage(long accountId, long aggreementId, string providerName, string signedByName, long legalEntityId) :base(signedByName, accountId)
+        {
+            AgreementId = aggreementId;
+            ProviderName = providerName;
+            LegalEntityId = legalEntityId;
+        }
+
+        public string ProviderName { get; }
```
robinoconnor: For consistency i'd move the public params above the constructors as this is how its been done elsewhere. i'll only mention it here rather than on each message

src/SFA.DAS.EAS.EmployerAccount.Events.UnitTests/MessageTests.cs
```diff
@@ -0,0 +1,41 @@
...
+        private const string TestStringProperty = "TestStringProperty";
+        private const long AccountId = 12345;
+        private const long TestLongProperty = 54321;
+
+        [Test]
+        public void WhenAMessageIsSerializedThenItCanBeDesialized()
+        {
+            var testClass = new TestMessage(SignedByName, AccountId, TestStringProperty, TestLongProperty);
+            var serialized = JsonConvert.SerializeObject(testClass);
+            var deserialized = JsonConvert.DeserializeObject<TestMessage>(serialized);
```
robinoconnor: If message is going to be serialisable we need to give it the correct serialisable interface as well.

src/SFA.DAS.EAS.Web/Orchestrators/EmployerAccountPayeOrchestrator.cs
```diff
@@ -182,6 +182,7 @@ public EmployerAccountPayeOrchestrator(IMediator mediator, ILog logger, ICookieS
             return new OrchestratorResponse<RemovePayeSchemeViewModel> { Data = model };
         }
 
+        //TODO: The message gets mutated by the method. Message needs to be made immutable
```
robinoconnor: Fair enough, makes sense. We originally didn't do this as it wasn't getting passed around everywhere but happy to make it immutable. Need to mention this to Michael just so he's aware of this as its a common features across projects

src/SFA.DAS.EAS.Web/Orchestrators/EmployerAgreementOrchestrator.cs
```diff
@@ -171,17 +171,12 @@ public EmployerAgreementOrchestrator(IMediator mediator, ILog logger, EmployerAp
         }
 
         public async Task<OrchestratorResponse> SignAgreement(string agreementid, string hashedId, string externalUserId,
-            DateTime signedDate)
+            DateTime signedDate, string companyName)
```
robinoconnor: might not be a company, and might be better to add the legal entity id so that if we have a charity and company with the same name we know which one it is.

src/SFA.DAS.EmployerAccounts.Events/Messages/ApprenticeChangesRequestedMessage.cs
```diff
@@ -0,0 +1,19 @@
+﻿
+
+namespace NuGet.Messages
+{
+    public class ApprenticeChangesRequestedMessage: Message
```
robinoconnor: These messages are in the commitments project. They should not be added in this project as they are not related to MA. I saw Nuget in the namespace and alarm bells started to ring. You'll need to change those messages in the commitments project.

src/SFA.DAS.EmployerApprenticeshipsService.Application.UnitTests/Commands/CreateAccountCommandTests/WhenICallCreateAccount.cs
```diff
@@ -36,9 +37,10 @@ public class WhenICallCreateAccount
         private Mock<IGenericEventFactory> _genericEventFactory;
         private Mock<IAccountEventFactory> _accountEventFactory;
         private Mock<IRefreshEmployerLevyService> _refreshEmployerLevyService;
+        private Mock<IMembershipRepository> _mockMembershipRepository;
         private const long ExpectedAccountId = 12343322;
         private const long ExpectedLegalEntityId = 2222;
-        private const string ExpectedHashString = "123ADF23";
+        private const string ExpectedHashString = "1234hash";
```
robinoconnor: just out of curiosity why was this value changed?

src/SFA.DAS.EmployerApprenticeshipsService.Application/Commands/AcceptInvitation/AcceptInvitationCommandHandler.cs
```diff
@@ -77,6 +82,13 @@ public class AcceptInvitationCommandHandler : AsyncRequestHandler<AcceptInvitati
                 RelatedEntities = new List<Entity> { new Entity { Id =$"Account Id [{existing.AccountId}], User Id [{user.Id}]", Type = "Membership" } },
                 AffectedEntity = new Entity { Type = "Invitation", Id = message.Id.ToString() }
             });
+
+            await PublishUserJoinedMessage(existing.AccountId, $"{user.FirstName} {user.LastName}");
```
robinoconnor: i'm wondering if the user model needs a full name property like the memberView has as this could be a very common format that would be needed elsewhere.

Also should we be using the userId as two people could have the same name?

src/SFA.DAS.EmployerApprenticeshipsService.Application/Commands/CreateAccount/CreateAccountCommandHandler.cs
```diff
@@ -64,32 +66,35 @@ public class CreateAccountCommandHandler : IAsyncRequestHandler<CreateAccountCom
...
 
-            var emprefs = message.PayeReference.Split(',');
 
             if (string.IsNullOrEmpty(message.OrganisationReferenceNumber))
             {
                 message.OrganisationReferenceNumber = Guid.NewGuid().ToString();
             }
 
-            var returnValue = await _accountRepository.CreateAccount(user.Id, message.OrganisationReferenceNumber, message.OrganisationName, message.OrganisationAddress, message.OrganisationDateOfInception, emprefs[0], message.AccessToken, message.RefreshToken, message.OrganisationStatus, message.EmployerRefName, (short)message.OrganisationType, message.PublicSectorDataSource, message.Sector);
+            var returnValue = await _accountRepository.CreateAccount(user.Id, message.OrganisationReferenceNumber, message.OrganisationName, message.OrganisationAddress, message.OrganisationDateOfInception, message.PayeReference, message.AccessToken, message.RefreshToken, message.OrganisationStatus, message.EmployerRefName, (short)message.OrganisationType, message.PublicSectorDataSource, message.Sector);
```
robinoconnor: thinking returnValue could be name something better now that we are changing the code

src/SFA.DAS.EmployerApprenticeshipsService.Application/Commands/CreateInvitation/CreateInvitationCommand.cs
```diff
@@ -8,8 +8,18 @@ public class CreateInvitationCommand : IAsyncRequest
     {
         public string ExternalUserId { get; set; }
         public string HashedAccountId { get; set; }
+
+        /// <summary>
+        /// The person being invited
+        /// </summary>
         public string Name { get; set; }
+        /// <summary>
+        /// The person being invited
```
robinoconnor: duplicated descriptions, need to put the correct values in

src/SFA.DAS.EmployerApprenticeshipsService.Application/Commands/CreateLegalEntity/CreateLegalEntityCommandHandler.cs
```diff
@@ -55,27 +59,24 @@ public class CreateLegalEntityCommandHandler : IAsyncRequestHandler<CreateLegalE
...
 
+            
+
             await CreateAuditEntries(owner, agreementView);
 
             await NotifyLegalEntityCreated(message.HashedAccountId, agreementView.LegalEntityId);
 
-            await PublishAgreementCreatedMessage(owner.AccountId, agreementView.LegalEntityId,
-                agreementView.Id);
+            await PublishAgreementCreatedMessage(_hashingService.DecodeValue(message.HashedAccountId),
```
robinoconnor: can we move the hashing service call outside of the method call so that it can be easily debugged later on?

src/SFA.DAS.EmployerApprenticeshipsService.Domain/Data/Entities/Account/Primitives/PayeAccountRef.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace SFA.DAS.EAS.Domain.Data.Entities.Account.Primitives
+{
```
robinoconnor: This should not be committed at the moment as it looks like its been removed from the project.

src/SFA.DAS.EmployerApprenticeshipsService.Domain/SFA.DAS.EAS.Domain.csproj
```diff
@@ -353,7 +353,9 @@
       <SubType>Designer</SubType>
     </None>
   </ItemGroup>
-  <ItemGroup />
+  <ItemGroup>
+    <Folder Include="Data\Entities\Account\Primitives\" />
```
robinoconnor: removed this as the folder is empty

src/SFA.DAS.EmployerApprenticeshipsService.Web.UnitTests/Orchestrators/EmployerAccountPayeOrchestratorTests/WhenRemovingAPayeScheme.cs
```diff
@@ -64,13 +64,19 @@ public void Arrange()
...
+                PayeRef = payeRef,
+                UserId = userRef
+            };
 
             //Act
             await _employerAccountPayeOrchestrator.RemoveSchemeFromAccount(model);
 
             //Assert
-            _mediator.Verify(x=>x.SendAsync(It.Is<RemovePayeFromAccountCommand>(c=>c.HashedAccountId.Equals(hashedId) && c.PayeRef.Equals(payeRef) && c.UserId.Equals(userRef))), Times.Once);
+            _mediator.Verify(x=>x.SendAsync(It.Is<RemovePayeFromAccountCommand>(c=>c.HashedAccountId.Equals(hashedId) && c.PayeRef.Equals(payeRef) && c.UserId.Equals(userRef) && c.CompanyName== SchemeName)), Times.Once);
```
robinoconnor: use .equals and not ==

src/SFA.DAS.EAS.EmployerAccount.Events.UnitTests/MessageTests.cs
TC1066: You mean Iserializable? All of our nessages will be serialised and deserialised. The reason for this test is to make sure the json serializer can do it's stuff when I've removed the setters.

src/SFA.DAS.EAS.Web/Orchestrators/EmployerAccountPayeOrchestrator.cs
TC1066: Michael is aware of the immutability features in the Activities work. It's partly at his request.

src/SFA.DAS.EAS.Web/Orchestrators/EmployerAgreementOrchestrator.cs
TC1066: We need to be specific. I am using it to construct a sentence. It could be argued that messages need to pass every actor as separate properties in case some processor needs it but I'm being Agile here.. just doing what I need to and being specific

src/SFA.DAS.EmployerAccounts.Events/Messages/AgreementSignedMessage.cs
TC1066: ok

src/SFA.DAS.EmployerAccounts.Events/Messages/ApprenticeChangesRequestedMessage.cs
TC1066: They are not in the solution. The files are on disk and not visible in VS. I'll delete them from disc.

TC1066: They were test files I used to make a realistic demo.

src/SFA.DAS.EmployerAccounts.Events/Messages/Message.cs
TC1066: There is already a message base class in shared messages. The current EAS messages do not use that, instead not have a common class. I've removed signed by name because it might not be used in all future files....having said that it's not being agile to look into the future...I should remove SignedByName when it becomes necessary to do so.

src/SFA.DAS.EmployerAccounts.Events/Messages/Message.cs
TC1066: It's not exteralUserId, it's the name of the person who pressed the button

src/SFA.DAS.EmployerAccounts.Events/Messages/Message.cs
TC1066: Comment above

src/SFA.DAS.EAS.EmployerAccount.Events.UnitTests/MessageTests.cs
robinoconnor: Yep ISerialiable, The tests are fine I just placed it there as a note as this was the place the serialisation was mentioned. Originally we did specifically state they were so now we are its worth just crossing the i's and dotting the T's. The tests are fine :)

src/SFA.DAS.EAS.Web/Orchestrators/EmployerAccountPayeOrchestrator.cs
robinoconnor: cool

src/SFA.DAS.EAS.Web/Orchestrators/EmployerAgreementOrchestrator.cs
robinoconnor: I'm confused as you name it as ProviderName in the message. So it can't be specific to a company if its called provider in the message. Are you using this as a way to persist the data so that if the legal entity changes we still have the historic name? Are these details not something we can attached to the agreement model itself if it needs to be persisted?
Otherwise we aren't we just sending all the data that everyone needs into the message? This would just bloat the message out.

src/SFA.DAS.EmployerAccounts.Events/Messages/ApprenticeChangesRequestedMessage.cs
robinoconnor: ok, that's fine. They just don't need to be committed that's all.

src/SFA.DAS.EmployerAccounts.Events/Messages/Message.cs
robinoconnor: externalUserId is the UserId of the person who's logged in and is used to get the name. If you need to persist the name then it may be useful to store it in the message however if we are processing these quickly that may not be an issue. The reason I was saying to would be better to store the ID is that you then can get further details like email address if you wanted without having to add any more code to the message. I think signedBy sounds wrong as well as it really the person who created it.

src/SFA.DAS.EmployerAccounts.Events/Messages/Message.cs
robinoconnor: The reason I mentioned about removing it was if it was moved to the shared project message project. You don't want to leak domain references in shared projects if you can help it. Same with Account ID. If you keep it in the MA domain these can remain but it might be useful to place the more generic stuff in the shared project message project as others might find it useful. We have done this with the message handling code so it's use it linked to us doing that with regards to the agile side.

src/SFA.DAS.EmployerAccounts.Events/Messages/Message.cs
robinoconnor: Just to add one more thing. When we designed the messaging system we didn't take into account the need for data to be persisted in the messages themselves. So sending the userId would then be picked up by activities and the user's name would then be persisted in the activity record itself. We could change this of course but that's why I mentioned it as the messages were not designed with this in mind. Events were designed like this which is why they cause some concern (especially with replaying external data sources).

robinoconnor: This could be the reason why it looks like i'm taking a different view as I'm seeing messages a short term signals that alert listeners and then those listeners collect the information they need to the IDs stored in the message. Its similar to the way Michael did the API calls for the RDS project.

src/SFA.DAS.EAS.Web/Orchestrators/EmployerAgreementOrchestrator.cs
robinoconnor: The reply above related to the message comments below about the persist nature of messages. This is why I'm raising the concern mainly as we don't use messages at present to persist data states. Not to my knowledge at least. which is why Ids are used to allow listeners of messages to collect all the data they need that relates to the message and then they can handle the persistence.

src/SFA.DAS.EmployerApprenticeshipsService.Application.UnitTests/Commands/CreateAccountCommandTests/WhenICallCreateAccount.cs
TC1066: I can only imagine that I had flip flopped from accountId to hashedAccountId and changed the string. It seems like an odd thing to do though admittedly and I can't see why I would have done it. A moment of complete madness.

src/SFA.DAS.EAS.Web/Orchestrators/EmployerAgreementOrchestrator.cs
TC1066: I've replaced companyName with organisationName in these messages. It is the case though that we need strings; names etc not ids. Michael confirms.

src/SFA.DAS.EmployerApprenticeshipsService.Application.UnitTests/Commands/CreateAccountCommandTests/WhenICallCreateAccount.cs
TC1066: I've set it back to what it was.

src/SFA.DAS.EmployerApprenticeshipsService.Application/Commands/AcceptInvitation/AcceptInvitationCommandHandler.cs
TC1066: Yep, done

src/SFA.DAS.EmployerApprenticeshipsService.Application/Commands/CreateAccount/CreateAccountCommandHandler.cs
TC1066: createAccountResult instead

src/SFA.DAS.EmployerApprenticeshipsService.Application/Commands/CreateInvitation/CreateInvitationCommand.cs
TC1066: Replaced by better property names

src/SFA.DAS.EmployerApprenticeshipsService.Application/Commands/CreateLegalEntity/CreateLegalEntityCommandHandler.cs
TC1066: OK

src/SFA.DAS.EmployerApprenticeshipsService.Domain/Data/Entities/Account/Primitives/PayeAccountRef.cs
TC1066: Deleted from disc

src/SFA.DAS.EmployerApprenticeshipsService.Domain/SFA.DAS.EAS.Domain.csproj
TC1066: ok

src/SFA.DAS.EmployerApprenticeshipsService.Web.UnitTests/Orchestrators/EmployerAccountPayeOrchestratorTests/WhenRemovingAPayeScheme.cs
TC1066: OK

src/SFA.DAS.EmployerApprenticeshipsService.Application/Commands/CreateAccount/CreateAccountCommandHandler.cs
```diff
@@ -117,42 +117,24 @@ public class CreateAccountCommandHandler : IAsyncRequestHandler<CreateAccountCom
...
         {
-            foreach (var empref in emprefs)
-            {
-                await _refreshEmployerLevyService.QueueRefreshLevyMessage(returnValue.AccountId, empref);
-            }
+            await _refreshEmployerLevyService.QueueRefreshLevyMessage(returnValue.AccountId, empref);
         }
 
-        private async Task PublishAddPayeSchemeMessage(IEnumerable<string> emprefs)
+        private async Task PublishAddPayeSchemeMessage(string empref, long accountId, string signedByName)
```
robinoconnor: Messages are still using signedByName parameters names. These need to be changed otherwise they will confuse people. Should be createdByName as not all messages have relations to signing agreement

src/SFA.DAS.EAS.Web/Orchestrators/EmployerAccountPayeOrchestrator.cs
TC1066: This is still marked as requiring changes

PR #1047, opened 2017-11-17
Changed the view unsigned agreements task link

robinoconnor:
- Now if you only have one unsigned agreement it will take you directly to it. Otherwise it will take you to the agreements list instead to review all agreements.

### DIFF
--- a/src/SFA.DAS.EAS.Web/Orchestrators/EmployerAgreementOrchestrator.cs
+++ b/src/SFA.DAS.EAS.Web/Orchestrators/EmployerAgreementOrchestrator.cs
@@ -85,7 +85,7 @@ namespace SFA.DAS.EAS.Web.Orchestrators
             return response;
         }
 
-        public async Task<OrchestratorResponse<EmployerAgreementListViewModel>> Get(string hashedId,
+        public virtual async Task<OrchestratorResponse<EmployerAgreementListViewModel>> Get(string hashedId,
             string externalUserId)
         {
             try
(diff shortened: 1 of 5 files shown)


src/SFA.DAS.EAS.Web/Orchestrators/EmployerAgreementOrchestrator.cs
```diff
@@ -85,7 +85,7 @@ public EmployerAgreementOrchestrator(IMediator mediator, ILog logger, EmployerAp
             return response;
         }
 
-        public async Task<OrchestratorResponse<EmployerAgreementListViewModel>> Get(string hashedId,
+        public virtual async Task<OrchestratorResponse<EmployerAgreementListViewModel>> Get(string hashedId,
```
arvinthseran: do we need make the method `virtual`?.

scottcowan: It looks like there's nothing overriding this method

robinoconnor: Its for mocking as we don't have interfaces for the orchestrators. The reason why we don't have interfaces for them was mainly political. Many devs didn't like orchestrators as there is a 1-1 relationship with controllers and so they didn't want to create the interfaces to highlight this issue. No completely sure its be useful tbh. I did mention the interfaces issues but was told to carry on not implementing them.

PR #1052, opened 2017-11-20
Aml 1240 activities

robinoconnor:


scottcowan:
currently the worker will only expect "paye scheme added" event, so I'd like to update it before we merge this

PR #1058, opened 2017-11-24
Halt levy declaration process if HMRC call fails.

MichaelYoung1981:
After retrying the call to get English Fraction data from HMRC 5 times the process swallows the exception and proceeds to get levy information which then uses the incorrect English Fraction amount. The solution to this is to halt the process whenever an exception occurs (after retrying 5 times). Ultimately this will dead letter the message so we will have to monitor the queues so we can resend the message.

### DIFF
--- a/src/SFA.DAS.EmployerApprenticeshipsService.Infrastructure/ExecutionPolicies/HmrcExecutionPolicy.cs
+++ b/src/SFA.DAS.EmployerApprenticeshipsService.Infrastructure/ExecutionPolicies/HmrcExecutionPolicy.cs
@@ -37,7 +37,7 @@ namespace SFA.DAS.EAS.Infrastructure.ExecutionPolicies
             {
                 _logger.Error(ex, $"Exceeded retry limit - {ex.Message}");
             }
-            return default(T);
+            throw ex;
         }
 
         private void OnRetryableFailure(Exception ex)


src/SFA.DAS.EmployerApprenticeshipsService.Infrastructure/ExecutionPolicies/HmrcExecutionPolicy.cs
```diff
@@ -37,7 +37,7 @@ protected override T OnException<T>(Exception ex)
             {
                 _logger.Error(ex, $"Exceeded retry limit - {ex.Message}");
             }
-            return default(T);
+            throw ex;
```
arvinthseran: Am i correct in saying that if you throw an error when you can not find an english fraction for that paye scheme then you can not add a paye scheme which has no english fraction. (application will always redirect you to There's been an error page and refuse to add the paye scheme)

MichaelYoung1981: Nope, the paye scheme is added and then a message is pushed on to a queue to get the English Fraction and Levy for the scheme as a background task.  If that process fails the user would be unaware and the scheme would still be added to their account.

PR #1060, opened 2017-11-28
Added content changes and unlocked tasks for new release

robinoconnor:


robinoconnor:
Change done so now merging

PR #1073, opened 2017-12-11
Added properties that will be used by activities

robinoconnor:


robinoconnor:
need to rewrite this to support what is in 1240 before it gets merged

PR #1079, opened 2017-12-14
Hide nav bar bug fix

robinoconnor:


src/SFA.DAS.EAS.Web/Extensions/DynamicExtentions.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System;
+
+namespace SFA.DAS.EAS.Web.Extensions
+{
+    public static class DynamicExtentions
```
scottcowan: Extentions > Extensions

src/SFA.DAS.EAS.Web/Extensions/DynamicExtentions.cs
```diff
@@ -0,0 +1,36 @@
...
+            if (hideNavString != null)
+            {
+                bool hideNavFlag;
+
+                if (bool.TryParse(hideNavString, out hideNavFlag))
+                {
+                    return hideNavFlag;
+                }
+
+                return !string.IsNullOrWhiteSpace(hideNavString);
```
scottcowan: could just be
```c#
return false;
```

PR #1085, opened 2017-12-19
AML-1240 Added activities

elliotritchie:


elliotritchie:
@robinoconnor Basically this PR is to show activities on the homepage and the activity page. I pretty much copied the way you'd setup tasks. These are the controller actions/Urls:
* https://github.com/SkillsFundingAgency/das-employerapprenticeshipsservice/blob/feature/AML-1240_paye_scheme_created/src/SFA.DAS.EAS.Web/Controllers/EmployerTeamController.cs#L30
* https://github.com/SkillsFundingAgency/das-employerapprenticeshipsservice/blob/feature/AML-1240_paye_scheme_created/src/SFA.DAS.EAS.Web/Controllers/EmployerTeamController.cs#L47

elliotritchie:
Yeah the defaults should be coming from here https://github.com/SkillsFundingAgency/das-employer-config/blob/master/das-activities/SFA.DAS.Activities.Client.json.

PR #1102, opened 2018-01-11
Aml 1819 just screen

DScuffle:
Can you review please

### DIFF
--- a/src/SFA.DAS.EAS.Web/Controllers/EmployerAccountTransactionsController.cs
+++ b/src/SFA.DAS.EAS.Web/Controllers/EmployerAccountTransactionsController.cs
@@ -33,17 +33,8 @@ namespace SFA.DAS.EAS.Web.Controllers
                 return RedirectToAction(ControllerConstants.IndexActionName, ControllerConstants.AccessDeniedControllerName);
             }
 
-            //transactionViewResult.Data.HashedAccountId = hashedAccountId;
             return View(transactionViewResult);
         }
-
-        //[Route("finance/downloadtransactions")]
-        //public async Task<ActionResult> TransactionsDownload(string hashedAccountId)
-        //{
-        //    await Task.Delay(11);
-
-        //    return View();
-        //}
 
         [Route("finance/{year}/{month}")]
         public async Task<ActionResult> TransactionsView(string hashedAccountId, int year, int month)
--- a/src/SFA.DAS.EAS.Web/Controllers/EmployerAccountTransactionsController.cs
+++ b/src/SFA.DAS.EAS.Web/Controllers/EmployerAccountTransactionsController.cs
@@ -15,25 +15,40 @@ namespace SFA.DAS.EAS.Web.Controllers
     {
         private readonly EmployerAccountTransactionsOrchestrator _accountTransactionsOrchestrator;
 
-        public EmployerAccountTransactionsController(IOwinWrapper owinWrapper, IFeatureToggle featureToggle, 
+        public EmployerAccountTransactionsController(IOwinWrapper owinWrapper, IFeatureToggle featureToggle,
             EmployerAccountTransactionsOrchestrator accountTransactionsOrchestrator, IMultiVariantTestingService multiVariantTestingService,
-            ICookieStorageService<FlashMessageViewModel> flashMessage) 
-            : base(owinWrapper, featureToggle,multiVariantTestingService,flashMessage)
+            ICookieStorageService<FlashMessageViewModel> flashMessage)
+            : base(owinWrapper, featureToggle, multiVariantTestingService, flashMessage)
         {
             _accountTransactionsOrchestrator = accountTransactionsOrchestrator;
         }
 
-        //TODO: Review this once we know how the URL should look
-        [Route("balance")]
+        [Route("finance")]
         public async Task<ActionResult> Index(string hashedAccountId)
         {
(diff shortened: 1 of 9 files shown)


src/SFA.DAS.EAS.Web/Views/EmployerAccountTransactions/Index.cshtml
```diff
@@ -1,239 +1,68 @@
...
-        </details>
-
-       
+        </div>
+    </div>
+</div>
+<div class="grid-row">
+    <div class="column-half">
+        <h2 class="heading-medium">
+            <a title="View transactions" href="@Url.Action("TransactionsView", "EmployerAccountTransactions", new {HashedAccountId = Model.Data.Account.HashedId, month = 0, year = 0})">
```
elliotritchie: This will likely append a `/0/0` path to the URL, is there an alternative?

src/SFA.DAS.EAS.Web/Controllers/EmployerAccountTransactionsController.cs
```diff
@@ -15,25 +15,40 @@ public class EmployerAccountTransactionsController : BaseController
...
 
-        //TODO: Review this once we know how the URL should look
-        [Route("balance")]
+        [Route("finance")]
         public async Task<ActionResult> Index(string hashedAccountId)
         {
-            return await Index(hashedAccountId, 0, 0);
+            var transactionViewResult = await _accountTransactionsOrchestrator.GetFinanceDashboardViewModel(hashedAccountId, 0, 0, OwinWrapper.GetClaimValue(ControllerConstants.SubClaimKeyName));
+
+            if (transactionViewResult.Data.Account == null)
```
elliotritchie: Will this code actually get called if there's a problem with the account hash or user validation?

src/SFA.DAS.EAS.Web/Controllers/EmployerAccountTransactionsController.cs
```diff
@@ -15,25 +15,40 @@ public class EmployerAccountTransactionsController : BaseController
...
+            }
+
+            //transactionViewResult.Data.HashedAccountId = hashedAccountId;
+            return View(transactionViewResult);
         }
-        
-        [Route("balance/{year}/{month}")]
-        public async Task<ActionResult> Index(string hashedAccountId, int year, int month)
+
+        //[Route("finance/downloadtransactions")]
```
scottcowan: is this feature toggled or just dead code?

src/SFA.DAS.EAS.Web/Views/EmployerAccountTransactions/Index.cshtml
scottcowan: if you leave them off I think it will use the default for whatever type that is on that action (probably int so 0)

src/SFA.DAS.EAS.Web/Views/EmployerAccountTransactions/TransactionsView.cshtml
```diff
@@ -0,0 +1,240 @@
...
+
+        <details class="feedback">
+            <summary><span class="summary">Help us improve this section</span></summary>
+            <div class="survey">
+                <script id="ss-embed-396029">
+                    (function (d, w) {
+                        var s, ss; ss = d.createElement('script');
+                        ss.type = 'text/javascript';
+                        ss.async = true;
+                        ss.src = ('https:' == d.location.protocol ? 'https://' : 'http://') + 'www.smartsurvey.co.uk/s/r/embed.aspx?i=350750&c=396029';
```
scottcowan: can we add this url to the config?

src/SFA.DAS.EAS.Web/Views/EmployerAccountTransactions/Index.cshtml
DScuffle: This does append a /0/0 that's, the "Your transaction" view requires this.

src/SFA.DAS.EAS.Web/Controllers/EmployerAccountTransactionsController.cs
DScuffle: This was Dead Code, it would have been used next feature, but I have deleted it for now.

src/SFA.DAS.EAS.Web/Views/EmployerAccountTransactions/TransactionsView.cshtml
DScuffle: This is the only usage, and I have no requirements over it, it was copy-paste existing code, so I think if this is required it is part of a bigger tech debt piece of work.

src/SFA.DAS.EAS.Web/Controllers/EmployerAccountTransactionsController.cs
DScuffle: Account is retrieved in the mediator and if not found empty view model returned

src/SFA.DAS.EAS.Web/Views/EmployerAccountTransactions/Index.cshtml
arvinthseran: we can use new `{HashedAccountId = Model.Data.Account.HashedId, month = DateTime.Now.Month, year = DateTime.Now.Year};` instead of 0/0 so application will redirects the page to the current month in current year

PR #1104, opened 2018-01-17
AML-2006 Added account ID to the homepage

elliotritchie:


src/SFA.DAS.EmployerApprenticeshipsService.Web.UnitTests/Controllers/EmployerTeamControllerTests/WhenIViewNextSteps.cs
```diff
@@ -34,10 +34,9 @@ public void Arrange()
 
             _controller = new EmployerTeamController(
                 _owinWrapper.Object,
-                _orchestrator.Object,
                 _featureToggle.Object,
                 _userViewTestingService.Object,
-                _flashMessage.Object);
+                _flashMessage.Object, _orchestrator.Object);
```
scottcowan: can you move the _orcestrator.Object to a new line

PR #1107, opened 2018-01-24
Aml 1901 download excel transaction file

robinoconnor:


src/SFA.DAS.EAS.Web/FileDownloads/Payments/CsvTransactionFormatter.cs
```diff
@@ -0,0 +1,80 @@
...
+            WriteRowsCsv(transactions, builder);
+			
+			var csvContent = builder.ToString();
+
+            return Encoding.UTF8.GetBytes(csvContent);
+        }
+
+        private static void WriteHeaderCsv(StringBuilder builder)
+        {
+            builder.AppendFormat("Transaction date,");
```
iangledhill: What is the benefit of calling AppendFormat(...) (here and in method below) over calling simple Append()? There are no args so the more complex AppendFormat(...) call path isn't needed.

src/SFA.DAS.EAS.Web/FileDownloads/Payments/CsvTransactionFormatter.cs
```diff
@@ -0,0 +1,80 @@
...
+            builder.AppendFormat("Total,");
+            builder.AppendFormat(Environment.NewLine);
+        }
+
+        private static void WriteRowsCsv(List<TransactionDownloadLine> transactions, StringBuilder builder)
+        {
+            foreach (var transaction in transactions)
+            {
+                builder.AppendFormat("{0},", transaction.DateCreated);
+                builder.AppendFormat("{0},", transaction.TransactionType);
```
iangledhill: Is string interpolation required here?  If not then could just use simpler Append(...) be used instead?

src/SFA.DAS.EAS.Web/FileDownloads/Payments/ExcelTransactionFormatter.cs
```diff
@@ -0,0 +1,62 @@
...
+        {
+            var excelRows = new List<string[]>();
+
+            excelRows.Add(GetHeaderRow());
+
+            foreach (var transaction in transactions)
+            {
+                excelRows.Add(new[]
+                {
+                    transaction.DateCreated.ToString(), transaction.TransactionType,
```
iangledhill: Using just ToString() means we cannot guarantee the format of the date (as the culture on the current thread will be used) - would it be better to use the invariant culture so we can guarantee format or perhaps use a fixed date format?

The same goes for elsewhere where dates use string interpolation. Int types ar eless of a problem but these should also be invariant.

src/SFA.DAS.EAS.Web/FileDownloads/Payments/ExcelTransactionFormatter.cs
```diff
@@ -0,0 +1,62 @@
...
+            };
+
+            var fileData = _excelService.CreateExcelFile(transactionData);
+
+            return fileData;
+        }
+
+        private string[] GetHeaderRow()
+        {
+            return new[]
```
iangledhill: I don't thinks this array is modified? If not could it be made private static so that a new array is not created for each request?

src/SFA.DAS.EAS.Web/FileDownloads/Payments/ITransactionFormatter.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System.Collections.Generic;
+using SFA.DAS.EAS.Domain.Models.Transaction;
+using SFA.DAS.EAS.Web.ViewModels;
+
+namespace SFA.DAS.EAS.Web.FileDownloads.Payments
+{
+    public interface ITransactionFormatter
+    {
+        byte[] GetFileData(List<TransactionDownloadLine> dataPayments);
```
iangledhill: If we define the List<T> as IEnumerable<T> would we make it more flexible? The concrete methods only do a foreach over the collections so don't need it to be a specific concrete type.

src/SFA.DAS.EAS.Web/Orchestrators/EmployerAccountTransactionsOrchestrator.cs
```diff
@@ -391,5 +402,45 @@ public EmployerAccountTransactionsOrchestrator(IMediator mediator, ICurrentDateT
...
+            //var data =
+            //    await
+            //        _mediator.SendAsync(new GetEmployerAccountPaymentsByDateQuery
+            //        {
+            //            HashedAccountId = hashedId,
+            //            ExternalUserId = externalUserId,
+            //            FromDate = fromDate,
+            //            ToDate = toDate
+            //        });
+
```
iangledhill: Should commented out code be removed?

src/SFA.DAS.EAS.Web/ViewModels/TransactionsDownloadResultViewModel.cs
```diff
@@ -1,7 +1,131 @@
...
+        }
+
+        public bool FromDateInFuture => ValidFromDate && DateInFuture(FromMonth, FromYear);
+
+        public bool ToDateInFuture => ValidToDate && DateInFuture(ToMonth, ToYear);
+
+        public DateTime FromDate()
+        {
+            if(!ValidFromMonth())
+                throw new ArgumentException("This object has an invalid FromMonth");
```
iangledhill: Is ArgumentException appropriate here?

src/SFA.DAS.EAS.Web/ViewModels/TransactionsDownloadResultViewModel.cs
```diff
@@ -1,7 +1,131 @@
...
+
+                return true;
+            }
+        }
+
+        public bool FromDateInFuture => ValidFromDate && DateInFuture(FromMonth, FromYear);
+
+        public bool ToDateInFuture => ValidToDate && DateInFuture(ToMonth, ToYear);
+
+        public DateTime FromDate()
```
iangledhill: Is this the best name? It is named like a property but is actually a method.

src/SFA.DAS.EAS.Web/FileDownloads/Payments/CsvTransactionFormatter.cs
robinoconnor: I didn't write this code as it was pulled in from David's branch so I'll ask him and change accordingly. Its showing up in mine as its currently not in master

src/SFA.DAS.EAS.Web/FileDownloads/Payments/ExcelTransactionFormatter.cs
robinoconnor: This one fell through the net, I'll update it now. We generally do use formats but I wouldn't be surprised if there are a few missing in the whole of the code. I was waiting to ask David what format he was using as his code wasn't finished but forgot to do this and didn't go back to change it. Given its being displayed we may have wanted to use currentculture to reflect what the server was setup for but I think a fixed format is better as we can then use a format excel supports easily.

src/SFA.DAS.EAS.Web/ViewModels/TransactionsDownloadResultViewModel.cs
```diff
@@ -1,7 +1,131 @@
...
+            if (!ValidFromYear())
+                throw new ArgumentException("This object has an invalid FromYear");
+
+            if (FromDateInFuture)
+                throw new ArgumentException("This FromYear is in the future");
+
+            return new DateTime(FromYear, FromMonth, 1);
+        }
+
+        public DateTime ToDate()
```
iangledhill: ToDate doesn't make it clear what is going on - it is the last second of the month that is being returned. Would it better to specify millisecond as well?

src/SFA.DAS.EAS.Web/FileDownloads/Payments/ExcelTransactionFormatter.cs
robinoconnor: yep correct, forgot to move it. done.

src/SFA.DAS.EAS.Web/FileDownloads/Payments/ITransactionFormatter.cs
robinoconnor: So I copied this for David's work again. i'm make the edit now as you are correct it would be better to use that.

src/SFA.DAS.EmployerApprenticeshipsService.Infrastructure/Services/ExcelService.cs
```diff
@@ -0,0 +1,77 @@
...
+
+namespace SFA.DAS.EAS.Infrastructure.Services
+{
+    public class ExcelService : IExcelService
+    {
+        public Dictionary<string, string[][]> ReadExcelFile(byte[] fileData)
+        {
+            var workbookData = new Dictionary<string, string[][]>();
+
+            using (var stream = new MemoryStream(fileData))
```
iangledhill: There is a lot of nesting going on here - could it be simplified?

src/SFA.DAS.EmployerApprenticeshipsService.Infrastructure/Services/ExcelService.cs
```diff
@@ -0,0 +1,77 @@
...
+                    foreach (var worksheetDetails in worksheets)
+                    {
+                        var worksheetName = worksheetDetails.Key;
+                        var worksheetData = worksheetDetails.Value;
+
+                        using (var worksheet = workbook.Worksheets.Add(worksheetName))
+                        {
+                            for (var rowIndex = 0; rowIndex < worksheetData.Length; rowIndex++)
+                            {
+                                //var rowData = worksheetData[rowIndex];
```
iangledhill: commented code?

robinoconnor: removed

src/SFA.DAS.EmployerApprenticeshipsService.Infrastructure/Services/ExcelService.cs
robinoconnor: it was change to increase performance as I had some issues with the old design that was a little less nested. The library i'm using only seems to allow cell by cell edits so by the nature of the importing it falls into a nested approach workbook -> worksheet -> row -> column. The outer nesting is handling the dispose aspects of the data handling objects. It fairly simple as it is but i could move some of the code to different methods if you think that would be useful?

src/SFA.DAS.EAS.Web/ViewModels/TransactionsDownloadResultViewModel.cs
robinoconnor: Pulled from the other branch but i'll take a look and modify as appropriate as its easier than pinging it back

src/SFA.DAS.EAS.Web/ViewModels/TransactionsDownloadResultViewModel.cs
robinoconnor: pulled from other branch but again i'll take a look

DScuffle:
Don't merge to master until we've pulled in Rajesh UI work

src/SFA.DAS.EAS.Web/FileDownloads/Payments/CsvTransactionFormatter.cs
robinoconnor: done

src/SFA.DAS.EAS.Web/ViewModels/TransactionsDownloadResultViewModel.cs
robinoconnor: i've changed this to an InvalidOperationException as this is the one Microsoft recommend for an object that is in an invalid state

src/SFA.DAS.EAS.Web/ViewModels/TransactionsDownloadResultViewModel.cs
robinoconnor: Given how the transactions work I doubt we need to go down to that level of accuracy. i'll update the method name though

DScuffle: Milli- seconds would not be good, but this is edge case

src/SFA.DAS.EmployerApprenticeshipsService.Application.UnitTests/Queries/GetEmployerAccountPaymentsByDate/WhenIGetEmployerAccountPaymentsByDate.cs
```diff
@@ -0,0 +1,79 @@
...
+        public override GetEmployerAccountPaymentsByDateQuery Query { get; set; }
+
+        public override GetEmployerAccountPaymentsByDateHandler RequestHandler { get; set; }
+
+        public override Mock<IValidator<GetEmployerAccountPaymentsByDateQuery>> RequestValidator { get; set; }
+
+        private const long ExpectedAccountId = 4564878;
+        private const string ExpectedUserId = "123dfv";
+
+        private readonly DateTime _fromDate = DateTime.Parse("01/01/2000");
```
iangledhill: This is dependent on the locale of the user running the tests - if we're building a new date should we use ParseExact or better in this case use one of the DateTime constructors?

src/SFA.DAS.EmployerApprenticeshipsService.Infrastructure/Services/ExcelService.cs
iangledhill: I personally would try to simplify it but if you're happy then that's cool - just leave it.

src/SFA.DAS.EmployerApprenticeshipsService.Application.UnitTests/Queries/GetEmployerAccountPaymentsByDate/WhenIGetEmployerAccountPaymentsByDate.cs
robinoconnor: I'll change this now, its another piece of code pulled over.

src/SFA.DAS.EAS.Web/FileDownloads/Payments/CsvTransactionFormatter.cs
```diff
@@ -0,0 +1,84 @@
...
+
+        private static void WriteRowsCsv(IEnumerable<TransactionDownloadLine> transactions, StringBuilder builder)
+        {
+            foreach (var transaction in transactions)
+            {
+                builder.Append($"{transaction.DateCreated:G},");
+                builder.Append($"{transaction.TransactionType},");
+                builder.Append($"{transaction.EmpRef},");
+                builder.Append($"{transaction.PeriodEnd},");
+                builder.Append($"{transaction.LevyDeclared.ToString(NumberFormatInfo.InvariantInfo)},");
```
iangledhill: I think the .ToString() is redundant - the string interpolation is going to call that anyway.

src/SFA.DAS.EAS.Employer_Financial.Database/StoredProcedures/GetAllTransactionDetailsForAccountByDate.sql
```diff
@@ -0,0 +1,122 @@
...
+AND TransactionType = 1
+AND DateCreated >= @fromDate 
+AND DateCreated <= @toDate
+
+
+INSERT INTO #output
+(
+	DateCreated,
+	TransactionType,
+	--EmpRef,
```
scottcowan: do you need the comments?

src/SFA.DAS.EAS.Web/ViewModels/DownloadFormatType.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace SFA.DAS.EAS.Web.ViewModels
+{
+    public enum DownloadFormatType
+    {
+        Csv,
```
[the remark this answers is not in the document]
DScuffle: Needs addressing in merge it has both entries on local repositories, but not through this page.

src/SFA.DAS.EmployerApprenticeshipsService.Domain/Models/Transaction/TransactionsDownloadEndDateMonthYearDateTime.cs
```diff
@@ -0,0 +1,25 @@
...
+        public override DateTime ToDateTime()
+        {
+            if (!ValidMonth)
+                throw new InvalidOperationException("This object has an invalid EndDate.Month");
+
+            if (!ValidYear)
+                throw new InvalidOperationException("This object has an invalid EndDate.Year");
+
+            // return 00:00 on the first day of the next month
+            return new DateTime(Year, Month + 1, 1);
```
elliotritchie: Will this still work if the month is December?

src/SFA.DAS.EmployerApprenticeshipsService.Application.UnitTests/Queries/GetTransactionsDownloadTests/WhenIGetTransactionsDownload.cs
```diff
@@ -0,0 +1,222 @@
...
+        private readonly TransactionsDownloadEndDateMonthYearDateTime _toDateNoResults =
+            new TransactionsDownloadEndDateMonthYearDateTime
+            {
+                Month = 1,
+                Year = 2000,
+            };
+
+        private GetTransactionsDownloadQueryHandler _handler;
+        private GetTransactionsDownloadQuery _query;
+        private GetTransactionsDownloadResponse _response;
```
elliotritchie: I think there's a unit test missing that asserts against the response being returned?

src/SFA.DAS.EmployerApprenticeshipsService.Application/Queries/GetTransactionsDownloadResultViewModel/GetTransactionsDownloadQueryHandler.cs
```diff
@@ -0,0 +1,97 @@
...
+            var membership = await _membershipRepository.GetCaller(message.AccountHashedId, _currentUser.ExternalUserId);
+
+            if (membership == null)
+            {
+                throw new UnauthorizedAccessException();
+            }
+
+            var validationResult = new ValidationResult();
+
+            if (message.StartDate.DateInFuture)
```
elliotritchie: The upper bound is validated but the lower bound isn't, is that ok? We could also validate that `StartDate` isn't greater than `EndDate`.

src/SFA.DAS.EAS.Web/ViewModels/TransactionDownloadViewModel.cs
```diff
@@ -0,0 +1,57 @@
...
+                    return Message.StartDate;
+                }
+
+                return null;
+            }
+            set
+            {
+                if (Message == null)
+                {
+                    Message = new GetTransactionsDownloadQuery();
```
elliotritchie: Because you're inheriting from `ViewModel<T>` this will be created automatically for you.

src/SFA.DAS.EAS.Web/Views/EmployerAccountTransactions/TransactionsDownload.cshtml
```diff
@@ -1,29 +1,95 @@
...
+        <div id="start-date-group" class="form-error-group form-group @(Html.IsValid(m => m.Message.StartDate) ? "" : "error")">
+            <span class="form-label-bold">Start date</span>
+            <span class="form-hint">For example, 5 2017</span>
+            <div id="StartDate" class="form-date">
+                @Html.ValidationMessageFor(m => m.Message.StartDate)
+                <div class="form-group form-group-month">
+                    <label for="StartDate.Month">
+                        Month
+                    </label>
+                    @Html.TextBox("StartDate.Month", Model.StartDate.Month, new { @class = "form-control length-limit", type = "number", maxlength = "2", min = "1", max = "12", aria_labelledby = "StartDate.Month" })
```
elliotritchie: I recommend using the expression based html helpers to keep the view strongly typed.

src/SFA.DAS.EAS.Web/Controllers/EmployerAccountTransactionsController.cs
```diff
@@ -41,10 +51,40 @@ public class EmployerAccountTransactionsController : BaseController
...
+            };
+
+            return View(viewModel);
+        }
+
+        [HttpPost]
+        [ValidateAntiForgeryToken]
+        [ValidateModelState]
+        [Route("finance/downloadtransactionsdata")]
+        public async Task<ActionResult> TransactionDownloadByDate(TransactionDownloadViewModel model)
```
elliotritchie: I recommend overloading your `HttpGet` method here to make use of the `ValidateModelState` attribute's post-redirect-get behavior.

src/SFA.DAS.EmployerApprenticeshipsService.Infrastructure/Data/TransactionRepository.cs
```diff
@@ -139,6 +140,35 @@ public TransactionRepository(LevyDeclarationProviderConfiguration configuration,
...
+
+                return await c.QueryAsync<TransactionDownloadLine>(
+                    sql: "[employer_financial].[GetAllTransactionDetailsForAccountByDate]",
+                    param: parameters,
+                    commandType: CommandType.StoredProcedure);
+            });
+
+            var hmrcDateService = new HmrcDateService();
+            var list = result.ToList();
+            foreach (var res in list)
```
elliotritchie: I think with the previous `.ToList()` that it's enumerating the results twice here.

elliotritchie:
@DScuffle I've made some changes, let me know what you think. I've taken your `TransactionsDownloadEndDateMonthYearDateTime` idea and made it more generic so that we can use it elsewhere if/when we need to bind to DateTime from the UI. Have a look at some of the outdated change requests too, my changes addressed some of them but not all.

src/SFA.DAS.EAS.Employer_Financial.Database/StoredProcedures/GetAllTransactionDetailsForAccountByDate.sql
```diff
@@ -0,0 +1,122 @@
+﻿CREATE PROCEDURE [employer_financial].[GetAllTransactionDetailsForAccountByDate]
```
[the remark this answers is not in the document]
DScuffle: Actioned

src/SFA.DAS.EAS.Employer_Financial.Database/StoredProcedures/GetAllTransactionDetailsForAccountByDate.sql
```diff
@@ -0,0 +1,122 @@
...
+	tl.Amount - LevyDeclared as TenPercentTopUp,
+	tl.Amount
+from [employer_financial].TransactionLine tl
+LEFT JOIN [employer_financial].[TransactionLineTypes] tlt on tlt.TransactionType = IIF(Amount >= 0, 1, 2)
+inner join [employer_financial].LevyDeclarationTopup ldt on ldt.SubmissionId = tl.SubmissionId
+inner join [employer_financial].LevyDeclaration ld on ld.submissionid = tl.submissionid
+where tl.AccountId = @accountId 
+AND tl.TransactionType in(1,2) 
+AND DateCreated >= @fromDate 
+AND DateCreated <= @toDate
```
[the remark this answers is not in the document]
DScuffle: Actioned

src/SFA.DAS.EmployerApprenticeshipsService.Infrastructure/Data/TransactionRepository.cs
DScuffle: Actioned

src/SFA.DAS.EAS.Web/FileDownloads/Payments/CsvTransactionFormatter.cs
DScuffle: Actioned

src/SFA.DAS.EAS.Employer_Financial.Database/StoredProcedures/GetAllTransactionDetailsForAccountByDate.sql
```diff
@@ -0,0 +1,122 @@
...
+	COALESCE((SUM(pays3.Amount) * -1), 0) as EmployerContribution,
+	COALESCE((SUM(pays2.Amount) * -1), 0) as GovermentContribution,
+	COALESCE((SUM(pays1.[Amount]) * -1),0) + COALESCE((SUM(pays2.[Amount]) * -1),0) + COALESCE((SUM(pays3.[Amount]) * -1),0)   as Total
+
+FROM [employer_financial].[Payment] p
+  inner join [employer_financial].[PaymentTransactionTypes] ptt on ptt.TransactionType =  p.TransactionType
+  inner join [employer_financial].[PaymentMetaData] meta ON p.PaymentMetaDataId = meta.Id
+  inner join (select PeriodEnd,AccountId,ukprn, EmpRef, TransactionDate, DateCreated, LevyDeclared, EnglishFraction from employer_financial.TransactionLine where DateCreated >= @fromDate AND 
+        DateCreated <= @toDate) transLine on transLine.AccountId = p.AccountId and transLine.PeriodEnd = p.PeriodEnd and transLine.Ukprn = p.Ukprn
+  left join [employer_financial].[Payment] pays1
```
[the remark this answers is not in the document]
DScuffle: Requires a pivot, this was copy paste from another stored procedure. To Discuss...

src/SFA.DAS.EmployerApprenticeshipsService.Application.UnitTests/Queries/GetTransactionsDownloadTests/WhenIGetTransactionsDownload.cs
DScuffle: Actioned

PR #1109, opened 2018-01-24
Changed the next link for start sign agreement

iangledhill:
The purpose of this is to change the target when a user has added an organisation and selected start sign agreement path.

The new target requires a hashed agreement id, which needs collecting in the confirm end point which can then be passed through as a hidden form field (similar to the existing organisation name).

The existing next link operates as a "new search" - I'm not sure of the distinction between a new search and a non-new-search. The latter is not (or does not appear to be) part of the journey concerned in the Jira so I've left the non-new-search alone.


src/SFA.DAS.EAS.Web/Controllers/OrganisationController.cs
```diff
@@ -84,7 +84,7 @@ public ActionResult OrganisationLegalAgreement(string hashedAccountId, Organisat
             AddFlashMessageToCookie(flashMessage);
             if (newSearch)
             {
-                return RedirectToAction(ControllerConstants.OrganisationAddedNextStepsSearchActionName, new { hashedAccountId, organisationName = name });
+                return RedirectToAction(ControllerConstants.OrganisationAddedNextStepsSearchActionName, new { hashedAccountId, organisationName = name, hashedAGreementId = response.Data.EmployerAgreement.HashedAgreementId });
```
robinoconnor: AGreement => Agreement . G needs to be lower case

PR #1112, opened 2018-01-25
Aml 2039 improve bad request logging

iangledhill:
This adds response logging for bad requests. This is intended specifically to help with the bad requests that we're getting from HMRC but will log all bad requests.

It was necessary to upgrade das nlogger packages across the solution to the same version.

### DIFF
--- a/src/SFA.DAS.EmployerApprenticeshipsService.Infrastructure/Services/HttpClientWrapper.cs
+++ b/src/SFA.DAS.EmployerApprenticeshipsService.Infrastructure/Services/HttpClientWrapper.cs
@@ -21,14 +21,10 @@ namespace SFA.DAS.EAS.Infrastructure.Services
 
         private readonly IHttpResponseLogger _httpResponseLogger;
 
-        public HttpClientWrapper(ILog logger)
+        public HttpClientWrapper(ILog logger, IHttpResponseLogger httpResponseLogger) 
         {
             _logger = logger;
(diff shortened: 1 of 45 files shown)


src/SFA.DAS.EmployerApprenticeshipsService.Infrastructure/Services/HttpClientWrapper.cs
```diff
@@ -7,23 +7,31 @@
...
     public class HttpClientWrapper : IHttpClientWrapper
     {
         public string AuthScheme { get; set; }
         public string BaseUrl { get; set; }
         public List<MediaTypeWithQualityHeaderValue> MediaTypeWithQualityHeaderValueList { get; set; }
         private readonly ILog _logger;
 
+        private readonly IHttpResponseLogger _httpResponseLogger;
+
         public HttpClientWrapper(ILog logger)
```
elliotritchie: If there are multiple constructors on a class then StructureMap's default behavior is to select the greediest, so I think we can remove this? Will it blow up without an `IHttpResponseLogger` if a 400 response is returned anyway?

iangledhill: Cheers.
Are rats! Good spot - I put this additional constructor to in there while I was trying to figure out the ioc/multiple nlogger issue and it stuck in.

PR #1116, opened 2018-01-30
AML-2060 Added activities toggle

elliotritchie:


src/SFA.DAS.EmployerApprenticeshipsService.Application/Queries/GetActivities/GetActivitiesQueryHandler.cs
```diff
@@ -0,0 +1,78 @@
...
+        {
+            _currentUser = currentUser;
+            _activitiesClient = activitiesClient;
+            _logger = logger;
+            _membershipRepository = membershipRepository;
+        }
+
+        public async Task<GetActivitiesResponse> Handle(GetActivitiesQuery message)
+        {
+            var membership = await _membershipRepository.GetCaller(message.HashedAccountId, _currentUser.ExternalUserId);
```
robinoconnor: I'm wondering if having a null/empty check on both hashedAccountId and ExternalUserId would be useful here just in case this handler is called from code that is not directly linked to and actions request?

robinoconnor:
I've approved the code but its worth checking and dealing with that comment beforehand if you feel it needs to be added. Given it won't affect the actual code and is a performance tweak its still good to go to master

src/SFA.DAS.EmployerApprenticeshipsService.Application/Queries/GetActivities/GetActivitiesQueryHandler.cs
elliotritchie: Nice catch, thanks. I'll add a handler into the Mediatr pipeline at some point that does it.

PR #1117, opened 2018-01-30
Aml 1381 add integration tests

iangledhill:
This is a new project and some supporting framework classes to support integration tests. The tests are run as real HTTP requests to an in-memory hosted web server. There are two integration tests to demonstrate how this is expected to work. It currently only supports GET but other verbs will be added next.

The tester can:
1. Verify that the request returned an acceptable status code
2. The expected controller was created 
3. Detect when the server has an unhandled exception (which will be treated as a fail)

The in-memory server will be configured using the same config as the actual API web server uses (so all filters etc). There are some slight alterations due to the fact HTTP context is not available (as this is an IIS hosting specific). Currently the effect of this is that the dependency resolver is not the same as the run time one (although the config is the same). As everything else seems to use HttpContextWrapper I think this should remain the only difference).

There is no mocking - so real data is used. I will investigate the practicality of removing this dependency on a persistent store.


src/SFA.DAS.EAS.Web.IntegrationTests/TestUtils/ApiIntegrationTester.cs
```diff
@@ -0,0 +1,196 @@
...
+            _exceptionHandler = new IntegrationTestExceptionHandler(existingExceptionHandler);
+
+            configuration.Services.Replace(typeof(IAssembliesResolver), new TestWebApiResolver<AccountLegalEntitiesController>());
+            configuration.Services.Replace(typeof(IExceptionHandler), _exceptionHandler);
+
+            StructuremapMvc.Start();
+            StructuremapWebApi.Start();
+            var container = ((StructureMapWebApiDependencyResolver)GlobalConfiguration.Configuration.DependencyResolver).Container;
+            var requestContextMock = new Mock<IRequestContext>();
+            container.Inject(requestContextMock.Object);
```
elliotritchie: Just a thought but could a StructureMap registry just for the `SFA.DAS.EAS.Account.API.IntegrationTests` project be used for any overrides?

PR #1122, opened 2018-02-01
Aml 1381 add integration test data builder

iangledhill:
Extended integration tester:

1. Added data building
2. Separated out call requirements from call response to avoid overuse of generics
3. Tidied up test util code into separate folders
4. Cleaned up some of the API Tester code (e.g. customisation of IOC and error handling)
5. Modified global exception handler to treat InvalidRequestException the same as ValidationException (to stop validation errors causing 500 when they should be 400)


src/SFA.DAS.EAS.Web.IntegrationTests/TestUtils/ApiTester/ApiIntegrationTester.cs
```diff
@@ -0,0 +1,263 @@
...
+            configuration.Services.Replace(typeof(IExceptionHandler), _exceptionHandler);
+
+            StructuremapMvc.Start();
+            StructuremapWebApi.Start();
+
+            var container = ((StructureMapWebApiDependencyResolver)GlobalConfiguration.Configuration.DependencyResolver).Container;
+
+            var requestContextMock = new Mock<IRequestContext>();
+
+            container.Configure(c =>
```
elliotritchie: I'd recommend a StructureMap registry specifically for the `SFA.DAS.EAS.Account.API.IntegrationTests` project. This will allow them to be defined in one place rather than across `ApiIntegrationTester` & `DbBuilderDependentRepositories` etc. You'll also be able to specify their lifetimes which should remove the need to define them as `Lazy`. Unless I'm missing something?

PR #1124, opened 2018-02-02
Aml 2063 multi declaration handling

robinoconnor:


src/SFA.DAS.EAS.Web/Orchestrators/EmployerAccountTransactionsOrchestrator.cs
```diff
@@ -314,6 +313,36 @@ public EmployerAccountTransactionsOrchestrator(IMediator mediator, ICurrentDateT
             };
         }
 
+        private static ICollection<TransactionLine> AggregateLevyTransactions(ICollection<TransactionLine> transactions)
+        {
+            if (transactions == null || !transactions.Any())
```
iangledhill: Why use .Any()? transactions is an ICollection and so has a .Count property, which is much more efficient.

src/SFA.DAS.EAS.Web/Orchestrators/EmployerAccountTransactionsOrchestrator.cs
```diff
@@ -314,6 +313,36 @@ public EmployerAccountTransactionsOrchestrator(IMediator mediator, ICurrentDateT
...
+                return new List<TransactionLine>();
+
+            if (transactions.Count(t => t.TransactionType == TransactionItemType.Declaration) < 2)
```
iangledhill: The use of count will enumerate all items in the collection - this doesn't seem necessary when all we're finding out is whether there is a single item that matches the condition. Whether there is 2 or 2000 doesn't matter.

iangledhill: Thinking about it again I wonder if moving the array creation at L325 to the start and then using levytransactions.length < 2 at L321 would make L321 more obvious about what it's doing? Not sure about this - depends on the probability of there being < 2 txns.

src/SFA.DAS.EAS.Web/Orchestrators/EmployerAccountTransactionsOrchestrator.cs
```diff
@@ -314,6 +313,36 @@ public EmployerAccountTransactionsOrchestrator(IMediator mediator, ICurrentDateT
...
+
+            //We haven't populated all the values here as the view doesn't support them yet
+            var levyTransaction = new LevyDeclarationTransactionLine
+            {
+                Description = levyTransactions.First().Description,
+                DateCreated = levyTransactions.Max(t => t.DateCreated),
+                Amount = levyTransactions.Sum(t => t.Amount),
+                TransactionType = TransactionItemType.Declaration,
+                TransactionDate = levyTransactions.Max(t => t.TransactionDate)
+            };
```
iangledhill: There are four enumerations on levyTransactions here - each requiring a new instance of an enumerator to be created. Code looks clean but could this be done more efficiently?

src/SFA.DAS.EmployerApprenticeshipsService.Web.UnitTests/Orchestrators/EmployerAccountTransactionOrchestratorTests/WhenIGetAccountTransactions.cs
```diff
@@ -117,22 +119,71 @@ public void Arrange()
...
+            var levyDeclaration =
+                actualTransactions?.SingleOrDefault(t => t.TransactionType == TransactionItemType.Declaration);
+
+            //Assert
+            Assert.IsNotNull(actualTransactions);
+            Assert.IsNotNull(levyDeclaration);
+            Assert.IsFalse(string.IsNullOrWhiteSpace(levyDeclaration.Description));
+            Assert.AreEqual(2, actualTransactions.Count);
+            Assert.AreEqual(levyTransactions.Sum(t => t.Amount), actualTransactions.Single(t => t.TransactionType == TransactionItemType.Declaration).Amount);
+        }
```
iangledhill: A single assert per test makes CI test failures dead easy to diagnose - if the test case contains many asserts (like here) then a failing CI test is harder to diagnose. Could this be mitigated by adding a message to the Asserts? Just thinking if CI fails we'll see the test that failed and a stack trace and a message like "expecting null" but we won't be able to tell which assert specifically failed.

src/SFA.DAS.EAS.Web/Orchestrators/EmployerAccountTransactionsOrchestrator.cs
robinoconnor: Thats me defaulting to IEnumerable practice. Wasn't sure if the count was pre calculated. i'll change this.

src/SFA.DAS.EAS.Web/Orchestrators/EmployerAccountTransactionsOrchestrator.cs
robinoconnor: so I could do this as a do while or even create an extension to do this i guess so that it stops searching after finding x occurrences. I originally just placed the count in there as it was the simplest way to the check for the rest of the code without having to create a new collection.

I'm guessing if there are less that 2 occurrences it would be more efficient to use count as you don't have the overhead of creating a new collection which would still have to enumerate over all of the items in the collection. It would however be more efficient to pre-create levy transactions if there were more than 1 occurrences.

I'll create an extension to take a predicate that can be used to work out if there are x occurrences in a more efficient way.

src/SFA.DAS.EAS.Web/Orchestrators/EmployerAccountTransactionsOrchestrator.cs
robinoconnor: This was mainly a code amount trade off. I could have created two variables and added logic to a single enumeration but thought at the time it wouldn't be a massive performance hit anyway. I'll add the new logic so we have the best performance.

src/SFA.DAS.EmployerApprenticeshipsService.Web.UnitTests/Orchestrators/EmployerAccountTransactionOrchestratorTests/WhenIGetAccountTransactions.cs
robinoconnor: This is something I've always been confused at in this project. I used to write tests with a single assert but others didn't so move that that approach. i'll split this up anyway

src/SFA.DAS.EAS.Web/Orchestrators/EmployerAccountTransactionsOrchestrator.cs
iangledhill: Where(<predicate>).Take(2) then would do it?

src/SFA.DAS.EAS.Web/Orchestrators/EmployerAccountTransactionsOrchestrator.cs
iangledhill: Individually no big deal, but what happens when this is scaled up? .NET mem alloc is good but creating objects increases pressure on the garbage collector increasing the frequency of L0 garbage collections - when the whole process stops - every single .NET thread is suspended for what can be quite a few seconds. 
Trade off - things that make code easier such as LINQ often come with a performance penalty. If it were just one loop here but there are four over the same collection.

src/SFA.DAS.EmployerApprenticeshipsService.Web.UnitTests/Orchestrators/EmployerAccountTransactionOrchestratorTests/WhenIGetAccountTransactions.cs
iangledhill: Personally I like one-test:one-assert approach because I've found it makes diagnosing CI failures easier and keeps the intention of each test crystal clear. Multiple asserts also tend to require more setup which makes it harder to tell what specific scenario is being tests. Appreciate not everyone's pov though. If we are going to use one-test:many-asserts then adding messages to assert makes it easier to tell which assert failed.

src/SFA.DAS.EAS.Web/Orchestrators/EmployerAccountTransactionsOrchestrator.cs
robinoconnor: where would still iterate over the whole collection first.

robinoconnor: i'll change this now

robinoconnor: OK after our chat I implemented the linq where.take(2) and then realise the extension I created was even more efficient for the specific task as take does guarantee you'll get two so you have to count them to make sure where as the extension I've added doesn't need to do this. I've at least been refreshed on the chained linq process however.

PR #1128, opened 2018-02-06
ER-211 - Migrate accounts api client to support net standard

sachpatel:
Have run the `RunBuild.bat` with no errors. 👍


leewadhams:
I think we can delete packages.config can't we?

Also should we set the minimum version of the Api.Types packages reference to be the latest one? Maybe we need to get the Major/Minor version updated on the types package.

leewadhams:
@elliotritchie This is the second half of the work to migrate the api client. Could you have a look through please.

elliotritchie:
We got some issues restoring NuGet packages for `SFA.DAS.EAS.Account.Api.Client`:

* `SFA.DAS.Account.Api.Types` can't be restored. It looks like `Build-57098` included the .NET Standard change from PR #1125 so not sure why Visual Studio is complaining.
* `Microsoft.Net.Http` is restored using the .NET Framework dll.

Do you get the same?

src/SFA.DAS.Account.Api.Client/SFA.DAS.EAS.Account.Api.Client.csproj
```diff
@@ -1,79 +1,25 @@
...
-    <Compile Include="AccountApiClient.cs" />
-    <Compile Include="AccountApiConfiguration.cs" />
-    <Compile Include="IAccountApiClient.cs" />
-    <Compile Include="IAccountApiConfiguration.cs" />
-    <Compile Include="Properties\AssemblyInfo.cs" />
-    <Compile Include="SecureHttpClient.cs" />
+    <PackageReference Include="System.Net.Http" Version="4.3.3" />
+    <PackageReference Include="Newtonsoft.Json" Version="10.0.3" />
+    <PackageReference Include="Microsoft.IdentityModel.Clients.ActiveDirectory" Version="3.16.0" />
+    <PackageReference Include="SFA.DAS.Account.Api.Types" Version="1.2.0" />
```
elliotritchie: Should this be `1.1.0.57098`?

src/SFA.DAS.Account.Api.Client/SFA.DAS.EAS.Account.Api.Client.csproj
```diff
@@ -1,79 +1,25 @@
...
   </ItemGroup>
   <ItemGroup>
-    <None Include="packages.config" />
-    <None Include="SFA.DAS.Account.Api.Client.nuspec" />
+    <ProjectReference Include="..\SFA.DAS.EAS.Account.Api.Types\SFA.DAS.EAS.Account.Api.Types.csproj" />
```
elliotritchie: It looks like the `SFA.DAS.Account.Api.Types` project & nuget package are both being referenced. When the `SFA.DAS.Account.Api.Client` nuget package is published will this cause any issues?

sachpatel: Yep, the package reference is redundant. Have removed.

PR #1136, opened 2018-02-13
AML-2085 & AML-2087 Update activities nuget

elliotritchie:


elliotritchie:
We'll create a new PR that will include multiple activity client/html fixes.

PR #1138, opened 2018-02-14
AML-2109 Updated feature toggle service to log the users external ID instead of their email

elliotritchie:


elliotritchie:
Completed as part of #1132 instead.

PR #1143, opened 2018-02-16
Aml 1907 display transfer allowance

robinoconnor:
This work is around the displaying and calculating of the transfer allowance in the transfer dashboard. Currently it is not in scope to work out how payments and transfers will reduce this value so it just works out the initial figure at the moment


src/SFA.DAS.EAS.Employer_Financial.Database/StoredProcedures/GetAccountTransferAllowance.sql
```diff
@@ -0,0 +1,30 @@
...
+				 @financialYearEndDate = DATEFROMPARTS(@currentYear, 4, 20)
+			END
+		ELSE
+			BEGIN
+				SELECT @financialYearStartDate = DATEFROMPARTS(@currentYear - 2, 4, 20),
+				 @financialYearEndDate = DATEFROMPARTS(@currentYear - 1, 4, 20)
+			END
+	END
+
+	SELECT SUM(Amount) * 0.1 FROM [employer_financial].[TransactionLine]
```
MichaelYoung1981: The allowance %age shouldn't be hard coded.

src/SFA.DAS.EAS.Employer_Financial.Database/StoredProcedures/GetAccountTransferAllowance.sql
```diff
@@ -0,0 +1,30 @@
+﻿CREATE PROCEDURE [employer_financial].[GetAccountTransferAllowance]
```
MichaelYoung1981: The transfer allowance needs to be calculated once a year and stored as late declarations and adjustments shouldn't be taken into consideration.

robinoconnor: But payments will affect this values so we are going to need to on demand unless we change it every time a payment transaction is added to the account?

MichaelYoung1981: Yes, following our discussion yesterday my comment was rubbish!

robinoconnor: i'm going to leave this issue as it is as we will cover this when we do the payments work. This code will not be released until then as its part of the feature. How we manage payments will be the main affect of how we cache/store pre calculated values.

robinoconnor:
I've sorted out the one issue from the code review so this should be good to go.

PR #1144, opened 2018-02-21
Add order by descending date on txn download

iangledhill:
Nothing else to add to the title.

### DIFF
--- a/src/SFA.DAS.EmployerApprenticeshipsService.Infrastructure/Data/TransactionRepository.cs
+++ b/src/SFA.DAS.EmployerApprenticeshipsService.Infrastructure/Data/TransactionRepository.cs
@@ -165,7 +165,7 @@ namespace SFA.DAS.EAS.Infrastructure.Data
                 }
             }
 
-            return transactionDownloadLines.ToList();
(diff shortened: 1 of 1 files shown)


src/SFA.DAS.EmployerApprenticeshipsService.Infrastructure/Data/TransactionRepository.cs
```diff
@@ -165,7 +165,7 @@ public TransactionRepository(LevyDeclarationProviderConfiguration configuration,
                 }
             }
 
-            return transactionDownloadLines.ToList();
+            return transactionDownloadLines.OrderByDescending(txn => txn.DateCreated).ToList();
```
elliotritchie: Could this be done in the `GetAllTransactionDetailsForAccountByDate` sproc instead? Not an issue if not, your call.

iangledhill: Yeah, did consider that but I don't think the sproc could know in advance what the required order should be. Also tend to favour getting the date out of SQL as fast as possible and leave presentation stuff to the consumer.

PR #1123, opened 2018-02-01
AML-2185 & AML-2186 Add provider apprenticeship and course details to paymentcreated event

abroodie:

PR #1145, opened 2018-02-21
Ci 576 ci 527 add new levy declaration event and enhance payment created event

abroodie:
This branch only contains the new/updated classes for the levy declaration and payment created events/messages.  

Projects that reference the SFA.DAS.Provider.Events.Api.Client nuget package have also been updated to reference the latest version.


elliotritchie:
Superseded by #1123.

PR #1147, opened 2018-02-26
Sink exception when fetching latest activities on dashboard

iangledhill:
This sinks exceptions when rendering the latest activities on the dashboard page.

### DIFF
--- a/src/SFA.DAS.EAS.Web/Controllers/ActivitiesController.cs
+++ b/src/SFA.DAS.EAS.Web/Controllers/ActivitiesController.cs
@@ -1,4 +1,5 @@
-﻿using System.Threading.Tasks;
+﻿using System;
+using System.Threading.Tasks;
 using System.Web.Mvc;
 using AutoMapper;
 using MediatR;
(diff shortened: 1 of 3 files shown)


src/SFA.DAS.EAS.Web/Controllers/ActivitiesController.cs
```diff
@@ -34,10 +35,17 @@ public ActivitiesController(IMapper mapper, IMediator mediator)
...
-            var model = _mapper.Map<LatestActivitiesViewModel>(response);
+            try
+            {
+                var response = _mediator.Send(query);
+                var model = _mapper.Map<LatestActivitiesViewModel>(response);
 
-            return PartialView(model);
+                return PartialView(model);
+            }
+            catch (Exception e)
```
elliotritchie: Is it worth logging `e`?

src/SFA.DAS.EAS.Web/Controllers/ActivitiesController.cs
```diff
@@ -34,10 +35,17 @@ public ActivitiesController(IMapper mapper, IMediator mediator)
         [Route("latest")]
         public ActionResult Latest(GetLatestActivitiesQuery query)
         {
-            var response = _mediator.Send(query);
-            var model = _mapper.Map<LatestActivitiesViewModel>(response);
+            try
```
elliotritchie: Is it worth adding a unit test for the try catch logic?

src/SFA.DAS.EAS.Web/Controllers/ActivitiesController.cs
iangledhill: Added

src/SFA.DAS.EAS.Web/Controllers/ActivitiesController.cs
iangledhill: Done

PR #1152, opened 2018-02-28
Aml 2137 deserialise feature config

iangledhill:
This uses new classes to represent the agreed logical layout and adds in a cache that optimises the look ups.


elliotritchie:
Merged into #1162.

PR #1159, opened 2018-03-07
Payment checker

robinoconnor:


src/SFA.DAS.EAS.DbMaintenance.WebJob/Jobs/PaymentIntegrityCheckerJob.cs
```diff
@@ -0,0 +1,74 @@
...
+            var accounts = await _employerAccountRepository.GetAllAccounts();
+
+            var periodEnds = await _levyRepository.GetAllPeriodEnds();
+
+            foreach (var periodEnd in periodEnds)
+            {
+                foreach (var account in accounts)
+                {
+                    var expectedPayments =
+                        (IEnumerable<Payment>)await _paymentService.GetAccountPayments(periodEnd.Id,
```
iangledhill: Is the cast required? The returned ICollection<T> is already an IEnumerable<T> (PaymentDetail is a subclass of Payment)
The return type can't be null (IEnumerable derived return types are not supposed to be null) and in this case the actual called code will never return null,  so the subsequent Elvis operators starting at L 49 aren't required.

src/SFA.DAS.EmployerApprenticeshipsService.Domain/Models/Payments/Payment.cs
```diff
@@ -26,6 +26,49 @@ public class Payment
...
+            if (EmployerAccountVersion != payment.EmployerAccountVersion) return false;
+            if (ApprenticeshipVersion != payment.ApprenticeshipVersion) return false;
+            if (FundingSource != payment.FundingSource) return false;
+            if (TransactionType != payment.TransactionType) return false;
+            if (Amount != payment.Amount) return false;
+            if (StandardCode != payment.StandardCode) return false;
+            if (FrameworkCode != payment.FrameworkCode) return false;
+            if (ProgrammeType != payment.ProgrammeType) return false;
+            if (PathwayCode != payment.PathwayCode) return false;
+            if (PathwayName != payment.PathwayName) return false;
```
iangledhill: The string comparisons here will be case sensitive - don't know if this is deliberate.

robinoconnor: Yep that should be fine as they should match completely.

PR #1162, opened 2018-03-09
Aml 2139 create pipeline strategy

robinoconnor:
> AML-2140 PR branch is based on this branch and supersedes this PR. Don't bother PR-ing this - do AML-2140 instead

There are two different sides to this development:
1. an authorisation that implements a kind-of chain-of-responsibility
2. an actual implementation of a link in that chain which checks the feature toggle

The IOperationAuthorisationService has a single method on it:
`IsOperationAuthorised(controllerName, actionName, membershipContext)`
This wraps the three parameters into an OperationContext and passes that to the IOperationAuthorisationHandler.CanAccessAsync() method. 

There are two implementations of this interface:
- OperationAuthorisation 
- FeatureToggleAuthorisationHandler

The former is a composite wrapper for all other implementations and implements the chain of responsibility. 
The second determines whether a toggled feature is available to the current user. It actually delegates all the work to the IFeatureToggleService service. 

The IFeatureToggleService uses json doc from table storage to get the list of versions and uses the OperationFilterAttribute to determine what end points are linked to what features. An end point may be linked to as many features as required.

The storage object and the mappings are cached and structured for quick access to allow the access checks to be efficient.


src/SFA.DAS.EmployerApprenticeshipsService.Infrastructure/Pipeline/Features/Sections/WhitelistPipelineSection.cs
```diff
@@ -0,0 +1,81 @@
...
+using SFA.DAS.EAS.Infrastructure.Services.FeatureToggle;
+using SFA.DAS.NLog.Logger;
+using System;
+using System.Linq;
+using System.Text.RegularExpressions;
+using System.Threading.Tasks;
+
+namespace SFA.DAS.EAS.Infrastructure.Pipeline.Features.Sections
+{
+    public class WhitelistPipelineSection :
```
elliotritchie: The handler should ideally be composed rather than inheriting its behaviour. It probably shouldn't know about the caching strategy either.

elliotritchie:
Merged into PR #1178.

PR #1167, opened 2018-03-14
AML-2191 Removes top-up data etc on non-levy-declarations. Adds Tests

DScuffle:
The GetLevyDeclaration (AndTopUp) database views will no longer have any non-declarations showing.

There are acceptance Tests added that check for zero and null declarations with NoPaymentForPeriod set to true. Tests that normal service is resumed after one of these declarations.

Changes to EmployerAccountTransactionsOrchestrator.cs  were made cause I needed defensive code in these tests.

N.B. Most of the feature updates were only because SpecFlow required it.


src/SFA.DAS.EAS.Web/Orchestrators/EmployerAccountTransactionsOrchestrator.cs
```diff
@@ -62,16 +62,22 @@ public EmployerAccountTransactionsOrchestrator(IMediator mediator, ICurrentDateT
...
                 Status = HttpStatusCode.OK,
                 Data = new TransactionLineViewModel<LevyDeclarationTransactionLine>
                 {
                     Amount = data.Total,
                     SubTransactions = data.Transactions,
-                    TransactionDate = data.Transactions.First().DateCreated
                 }
             };
+
+            if (data.Transactions.Any())
```
elliotritchie: This looks like `outputData.Data.TransactionDate` might not be set and so will default to `DateTime.MinValue`. Is that ok?

DScuffle: Good spot, I have changed it to return a null data set if we have no declarations and reverted that block of code to the original. The acceptance test in turn now check that if we have a null dataset we expected zero levy and topup.

elliotritchie: Will returning a null view model break the views?

DScuffle: No more than they already breaks! You should always be clicking through a "Parent" link to get to that screen, so should never find yourself there if there is no record. If you do fudge the Url to get there it current errors with sequence contains no element anyway!

I have tested in the UI and both payment and Levy drill downs continue to work when screen links used.

PR #1177, opened 2018-03-20
AML-2194 Migrated SFS.DAS.EAS.Support

IainStevenson:
Also updated WhenTransformingConfigFiles to cope with Support config


IainStevenson:
I had deferred that decision to avoid it being a blocker on moving the code. We need to discuss it prior to the DEVOPS final changes in ASCS-224

PR #1178, opened 2018-03-23
Aml 2140 add agreement check auth handler

iangledhill:
This does several things
1. adds a new handler to deal with linking agreements to features
2. removes support for multiple features per end point (i.e. a single end point may be enabled by several features)
3. adds support for version agreements


elliotritchie:
Superseded by PR #1185.

PR #1183, opened 2018-03-27
Aml 2141 invalidate json cache

iangledhill:
The problem this was trying to solve is:
Ensure that if I sign a new agreement that entitles me to new features I can access them immediately after signing. Any caches that are used to speed up the agreement based access checks must be immediately invalidated.

Previously there were three caches: 2 x ICache (InMemoryCache and RedisCache) and ICacheProvider (InProcessCacheProvider). The InMemoryCache and InProcessCacheProvider are functionally equivelant, so ICacheProvider has been removed and its duties moved to InMemoryCache. The RedisCache was not used. The two ICache implementations have been separated into IInProcessCache and IDistributedCache (as the requirements from a consumer pov are completely different and they are not interchangeable).


elliotritchie:
Merged into PR #1178.

PR #1192, opened 2018-04-03
Aml 2238 all agreements must be signed

iangledhill:
This uses the latest agreement signed by **all** legal entities that are associated with an account as the applicable agreement. Previously, an agreement was applicable when **any** legal entity signed the agreement.


elliotritchie:
Cherry picked into PR #1185.

PR #1193, opened 2018-04-04
AML-2174 Add delete confirmation

iangledhill:


src/SFA.DAS.EAS.Web/Controllers/TransferConnectionInvitationsController.cs
```diff
@@ -220,14 +220,41 @@ public ActionResult Rejected(RejectedTransferConnectionInvitationViewModel model
...
-                    return RedirectToAction("Index", "Transfers");
+                    return RedirectToAction("Deleted");
                 case "GoToTransfersPage":
                     return RedirectToAction("Index", "Transfers");
                 default:
                     throw new ArgumentOutOfRangeException(nameof(model.Choice));
             }
         }
 
+        [HttpNotFoundForNullModel]
```
elliotritchie: The HttpNotFoundForNullModel attribute shouldn't be needed here.

PR #1196, opened 2018-04-07
AML-2237 Updated transfers dashboard to request transfer requests from commitments api

elliotritchie:


src/SFA.DAS.EAS.Account.Worker.UnitTests/Foo.cs
```diff
@@ -0,0 +1,13 @@
+﻿using NUnit.Framework;
```
robinoconnor: Is this needed? Looks like its a placeholder that never got replaced.

src/SFA.DAS.EAS.Api/Web.config
```diff
@@ -194,7 +194,7 @@
       </dependentAssembly>
       <dependentAssembly>
         <assemblyIdentity name="Microsoft.IdentityModel.Clients.ActiveDirectory" publicKeyToken="31bf3856ad364e35" culture="neutral" />
-        <bindingRedirect oldVersion="0.0.0.0-3.16.0.14" newVersion="3.16.0.14" />
+        <bindingRedirect oldVersion="0.0.0.0-3.13.8.999" newVersion="3.13.8.999" />
```
robinoconnor: What's the reason for the downgrading of versions?

src/SFA.DAS.EAS.DbMaintenance.UnitTests/app.config
```diff
@@ -28,7 +28,7 @@
       </dependentAssembly>
       <dependentAssembly>
         <assemblyIdentity name="Microsoft.IdentityModel.Clients.ActiveDirectory" publicKeyToken="31bf3856ad364e35" culture="neutral" />
-        <bindingRedirect oldVersion="0.0.0.0-3.16.0.14" newVersion="3.16.0.14" />
+        <bindingRedirect oldVersion="0.0.0.0-3.13.8.999" newVersion="3.13.8.999" />
```
robinoconnor: Again why are we downgrading the version?

src/SFA.DAS.EAS.Web/Controllers/ActivitiesController.cs
```diff
@@ -51,7 +50,7 @@ public ActionResult Latest(GetLatestActivitiesQuery query)
             {
                 _logger.Warn(ex, "Failed to get the latest activities.");
 
-                return Content(ControllerConstants.ActivitiesUnavailableMessage);
+                return Content("Activities are currently unavailable.");
```
robinoconnor: Are we not using constants anymore?

src/SFA.DAS.EAS.Account.Worker.UnitTests/Foo.cs
elliotritchie: Yeah, if there's no tests in the assembly the build fails so I added a placeholder. I'd delete the project but it's being re-purposed by Ian on his branch.

src/SFA.DAS.EAS.Api/Web.config
elliotritchie: Fixed

src/SFA.DAS.EAS.DbMaintenance.UnitTests/app.config
elliotritchie: Fixed

src/SFA.DAS.EAS.Web/Controllers/ActivitiesController.cs
elliotritchie: Fixed

PR #1199, opened 2018-04-11
AML-2242 AML-2409 Add breakdown of levy by date rather than one per month

iangledhill:
Previously the levy declarations were aggregated to one per month, dated as the last levy declaration in that month. 
This is not the required behaviour.
A separate line for each day on which levy was paid should be shown. 

Also fixed a minor, unrelated bug which has the "balance at" date set to the system date even though the presented data related to a specific year and month. The date is now set to the month end of the relevant month.


elliotritchie:
@iangledhill @robinoconnor AML-2242 has been closed, should it have been?

PR #1201, opened 2018-04-11
Aml 2097 duplicate submission ids from hmrc

DScuffle:
In this PR we are going to:
Copy the LevyDeclaration data to a newly created table.
Delete any duplicated submisisonIds from the LevyDeclaration table.
Apply a Unique Index to the LevyDeclaration for EmpRef and SubmissionId.

Moving forward Importing Levy Declarations will strip out any duplicate submission Ids from the batch being processed so only one is added. ~If any duplicates were removed it will log this as info.

### DIFF
--- a/src/SFA.DAS.EAS.Employer_Financial.Database/SFA.DAS.EAS.Employer_Financial.Database.sqlproj
+++ b/src/SFA.DAS.EAS.Employer_Financial.Database/SFA.DAS.EAS.Employer_Financial.Database.sqlproj
@@ -144,6 +144,7 @@
     <None Include="Scripts\Manual\AddNonTransferPayments.sql" />
     <None Include="Scripts\Manual\AddTransferLevy.sql" />
     <None Include="Scripts\Manual\CreateTransferPayments.sql" />
+    <None Include="Scripts\PreDeployment\RemoveDuplicateLevyDeclarations.sql" />
   </ItemGroup>
   <ItemGroup>
     <None Include="Database.publish.xml" />
--- a/src/SFA.DAS.EmployerApprenticeshipsService.Application/Commands/RefreshEmployerLevyData/RefreshEmployerLevyDataCommandHandler.cs
+++ b/src/SFA.DAS.EmployerApprenticeshipsService.Application/Commands/RefreshEmployerLevyData/RefreshEmployerLevyDataCommandHandler.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Text;
 using System.Threading.Tasks;
 using MediatR;
 using SFA.DAS.EAS.Application.Commands.PublishGenericEvent;
@@ -9,10 +10,12 @@ using SFA.DAS.EAS.Application.Exceptions;
 using SFA.DAS.EAS.Application.Factories;
 using SFA.DAS.EAS.Application.Validation;
 using SFA.DAS.EAS.Domain.Data.Repositories;
+using SFA.DAS.EAS.Domain.Extensions;
 using SFA.DAS.EAS.Domain.Interfaces;
 using SFA.DAS.EAS.Domain.Models.HmrcLevy;
 using SFA.DAS.EAS.Domain.Models.Levy;
 using SFA.DAS.HashingService;
+using SFA.DAS.NLog.Logger;
 
 namespace SFA.DAS.EAS.Application.Commands.RefreshEmployerLevyData
 {
@@ -25,9 +28,10 @@ namespace SFA.DAS.EAS.Application.Commands.RefreshEmployerLevyData
         private readonly ILevyEventFactory _levyEventFactory;
         private readonly IGenericEventFactory _genericEventFactory;
         private readonly IHashingService _hashingService;
+        private readonly ILog _logger;
 
         public RefreshEmployerLevyDataCommandHandler(IValidator<RefreshEmployerLevyDataCommand> validator, IDasLevyRepository dasLevyRepository, IMediator mediator, IHmrcDateService hmrcDateService,
-            ILevyEventFactory levyEventFactory, IGenericEventFactory genericEventFactory, IHashingService hashingService)
+            ILevyEventFactory levyEventFactory, IGenericEventFactory genericEventFactory, IHashingService hashingService, ILog logger)
         {
             _validator = validator;
             _dasLevyRepository = dasLevyRepository;
@@ -36,6 +40,7 @@ namespace SFA.DAS.EAS.Application.Commands.RefreshEmployerLevyData
             _levyEventFactory = levyEventFactory;
             _genericEventFactory = genericEventFactory;
             _hashingService = hashingService;
+            _logger = logger;
         }
 
         protected override async Task HandleCore(RefreshEmployerLevyDataCommand message)
@@ -53,6 +58,8 @@ namespace SFA.DAS.EAS.Application.Commands.RefreshEmployerLevyData
             foreach (var employerLevyData in message.EmployerLevyData)
             {
                 var declarations = employerLevyData.Declarations.Declarations.OrderBy(c => c.SubmissionDate).ToArray();
+
+                declarations = FilterDuplicateHmrcDeclarations(employerLevyData.EmpRef, declarations);
 
                 declarations = await FilterActiveDeclarations(employerLevyData, declarations);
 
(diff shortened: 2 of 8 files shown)


src/SFA.DAS.EAS.Employer_Financial.Database/SFA.DAS.EAS.Employer_Financial.Database.sqlproj
```diff
@@ -122,15 +124,17 @@
...
     <PostDeploy Include="SeedData\PostDeploymentRules.sql" />
     <None Include="SeedData\EnglishFraction.sql" />
   </ItemGroup>
   <ItemGroup>
-    <PreDeploy Include="Script.PreDeployment1.sql" />
+    <RefactorLog Include="SFA.DAS.EAS.Levy.Database.refactorlog" />
   </ItemGroup>
   <ItemGroup>
-    <RefactorLog Include="SFA.DAS.EAS.Levy.Database.refactorlog" />
+    <PreDeploy Include="Scripts\PreDeployment\PreDeployment1.sql" />
```
MichaelYoung1981: This file needs a considerably better name.

DScuffle:
In Order to Test this PR I have executed the SQL below to insert duplicate levy declarations and run the deploy:

I got a new Table called "LevyDeclarationNonUnique" appear with all my original Levy Declarations copied to it.

Duplicate declarations were removed from the "LevyDeclaration" table

Subsequent Deploys won't attempt to copy the table if this  "LevyDeclarationNonUnique" exists.

Every Subsequent Deploy **WILL** attempt to remove any duplicates (duplicates can never happen at this point as we will now have an Unique index) but should never make any further changes.



If the test is re-run you must delete the "LevyDeclarationNonUnique" table.



DECLARE @accountId int
DECLARE @empref varchar(50)

SET @accountId = 4
SET @empref = '444/AA00001'

INSERT INTO employer_financial.levydeclaration (AccountId,empref,levydueytd,levyallowanceforyear,submissiondate,submissionid,payrollyear,payrollmonth,createddate,hmrcsubmissionid)
VALUES (@accountId,    @empref,    1000,    1500.0000,    '2017-06-18 07:12:28.060',    1234567892,    '17-18',    2,    '2017-05-02 07:12:28.060',    123452)

INSERT INTO employer_financial.levydeclaration (AccountId,empref,levydueytd,levyallowanceforyear,submissiondate,submissionid,payrollyear,payrollmonth,createddate,hmrcsubmissionid)
VALUES (@accountId,    @empref,    1000,    1500.0000,    '2017-06-18 07:12:28.060',    1234567892,    '17-18',    2,    '2017-05-02 07:12:28.200',    123452)

src/SFA.DAS.EmployerApprenticeshipsService.Application/Commands/RefreshEmployerLevyData/RefreshEmployerLevyDataCommandHandler.cs
```diff
@@ -95,6 +100,24 @@ private static void ProcessNoPaymentForPeriodDeclarations(IEnumerable<DasDeclara
...
+        {
+            var dasDeclarations = declarations as DasDeclaration[] ?? declarations.ToArray();
+
+            var duplicateIds = dasDeclarations.GroupBy(d => d.SubmissionId).Where(g => g.Count() > 1)
+                .Select(s => s.First().SubmissionId).ToList();
+
+            duplicateIds.ForEach(submissionId =>
+                _logger.Info($"PayeScheme '{empRef}' has a duplicate submission id from Hmrc = '{submissionId}'"));
+
+            return dasDeclarations.GroupBy(x => x.SubmissionId).Select(s => s.First()).ToArray();
```
robinoconnor: Can you not use DistinctBy(x => x.submissionId)?

src/SFA.DAS.EAS.Employer_Financial.Database/Scripts/PreDeployment/AddLevyDeclarationCreatedDateDefault.sql
```diff
@@ -0,0 +1,22 @@
+﻿-- This script was created as there was already a fixup script that had not fully worked in Production
```
elliotritchie: At first glance I can't tell what this script is doing. If it's adding a default value to a column can this be added to the table's definition instead of a predeployment script?

DScuffle:
Elliott, with regard to the deployment script that adds a default (on top of setting it in the table declaration)....

From my test runs, adding it as a default onto the table is not enough, it does not get added when deploying against an existing database (not with my tooling anyway).

The script I added will mean it get's explicitly added to production when deployed, going forward all databases should get the default and this script can be removed.

src/SFA.DAS.EmployerApprenticeshipsService.Application/Commands/RefreshEmployerLevyData/RefreshEmployerLevyDataCommandHandler.cs
```diff
@@ -75,6 +81,23 @@ public class RefreshEmployerLevyDataCommandHandler : AsyncRequestHandler<Refresh
...
+        /// If there are any Submissions from Hmrc that have the same submission Id, we should discard all 
+        /// but the first. 
+        /// </summary> 
+        private DasDeclaration[] FilterDuplicateHmrcDeclarations(string empRef,
+            DasDeclaration[] declarations)
+        { 
+            var duplicateIds = declarations.GroupBy(d => d.SubmissionId).Where(g => g.Count() > 1)
+                .Select(s => s.First().SubmissionId).ToList(); 
+ 
+            duplicateIds.ForEach(submissionId =>
```
iangledhill: Do you think the message might make diagnosis easier if it was a single message containing the submission ids appended together into a csv string, so that we can see all the duplicated IDs in one place?

src/SFA.DAS.EmployerApprenticeshipsService.Application/Commands/RefreshEmployerLevyData/RefreshEmployerLevyDataCommandHandler.cs
```diff
@@ -75,6 +81,23 @@ public class RefreshEmployerLevyDataCommandHandler : AsyncRequestHandler<Refresh
...
+            DasDeclaration[] declarations)
+        { 
+            var duplicateIds = declarations.GroupBy(d => d.SubmissionId).Where(g => g.Count() > 1)
+                .Select(s => s.First().SubmissionId).ToList(); 
+ 
+            duplicateIds.ForEach(submissionId => 
+                _logger.Info($"PayeScheme '{empRef}' has a duplicate submission id from Hmrc = '{submissionId}'")); 
+
+
+            return declarations.DistinctBy(x => x.SubmissionId).ToArray();
```
iangledhill: This will take the first submission id from a group, whatever that happens to be. Is they're all have identical values that there is no difference, but what happens if they're different - is this still an acceptable way to select the row to retain?

src/SFA.DAS.EmployerApprenticeshipsService.Application/Commands/RefreshEmployerLevyData/RefreshEmployerLevyDataCommandHandler.cs
```diff
@@ -75,6 +82,25 @@ public class RefreshEmployerLevyDataCommandHandler : AsyncRequestHandler<Refresh
...
+        /// If there are any Submissions from Hmrc that have the same submission Id, we should discard all 
+        /// but the first. 
+        /// </summary> 
+        private DasDeclaration[] FilterDuplicateHmrcDeclarations(string empRef,
+            DasDeclaration[] declarations)
+        { 
+            var duplicateIds = declarations.GroupBy(d => d.SubmissionId).Where(g => g.Count() > 1)
+                .Select(s => s.First().SubmissionId).ToList();
+
+            var submissionIds = new StringBuilder() ;
```
elliotritchie: You could try this instead, it won't add the trailing comma & space `var submissionIds = string.Join(", ", duplicateIds)`.

src/SFA.DAS.EmployerApprenticeshipsService.Application/Commands/RefreshEmployerLevyData/RefreshEmployerLevyDataCommandHandler.cs
```diff
@@ -75,6 +82,25 @@ public class RefreshEmployerLevyDataCommandHandler : AsyncRequestHandler<Refresh
...
+            duplicateIds.ForEach(submissionId =>
+                submissionIds.AppendFormat("{0}, ", submissionId));
+
+            _logger.Info($"PayeScheme '{empRef}' has duplicate submission id(s) from Hmrc = {submissionIds}");
```
elliotritchie: It's might be worth adding single quotes around the parameter i.e. `'{submissionIds}'`.

elliotritchie:
Merged into #1268.

PR #1204, opened 2018-04-13
AML-2261 AML-2340 AML-2341 Payments import refactor

elliotritchie:


elliotritchie:
Also included as part of these changes: https://github.com/SkillsFundingAgency/das-shared-packages/pull/217.

src/AzureTopicConfigurationStrucuture.json
```diff
@@ -68,9 +68,5 @@
   {
     "topicName": "user_joined",
     "subscription": [ "Activity_UserJoinedMessageProcessor" ]
-  },
-  {
-    "topicName": "payment_created",
-    "subscription": [ "Activity_PaymentCreatedMessageProcessor" ]
```
MichaelYoung1981: If we remove this topic and, as unlikely as it is, build a new environment the topic won't exist and the publication of messages will throw an exception.

src/SFA.DAS.EmployerApprenticeshipsService.Infrastructure/Data/DasLevyRepository.cs
```diff
@@ -251,7 +252,7 @@ public DasLevyRepository(LevyDeclarationProviderConfiguration configuration, ILo
             return result;
         }
 
-        public async Task CreatePaymentData(IEnumerable<PaymentDetails> payments)
+        public async Task CreatePayments(IEnumerable<PaymentDetails> payments)
```
MichaelYoung1981: This method could do with a little bit of love.  Maybe extract out a few private methods?

elliotritchie: Done

src/AzureTopicConfigurationStrucuture.json
elliotritchie: Removing the config will prevent the subscription from being created if devops delete it. We can leave the topic there though. That should work?

PR #1212, opened 2018-04-23
AML-1946-2107-Transfer-Invitation-Response-Notification

DScuffle:
Has 3 Email templates put onto Gov Notify as well.

Notification link has been left as comment deliberately we need a strategy for this navigation to specific pages.

**REQUIRES DEPLOYMENT OF das-Notifications with config build above 76931.**


src/SFA.DAS.EAS.MessageHandlers.Worker/EventHandlers/ApprovedTransferConnectionInvitationEventHandler.cs
```diff
@@ -0,0 +1,79 @@
...
+        {
+            _db = dbContext;
+            _notificationsApi = notificationsApi;
+            _employerApprenticeshipsServiceConfiguration = employerApprenticeshipsServiceConfiguration;
+            _log = log;
+        }
+
+        protected override async Task ProcessMessage(ApprovedTransferConnectionInvitationEvent messageContent)
+        {
+            var users = _db.Users
```
elliotritchie: It looks like this query is used in a few places, could it be extracted out into an extension method hanging off `IQueryable<User>`?

src/SFA.DAS.EAS.MessageHandlers.Worker/NLog.config
```diff
@@ -0,0 +1,41 @@
+<?xml version="1.0" encoding="utf-8" ?>
```
elliotritchie: Is this file needed?

src/SFA.DAS.EAS.MessageHandlers.Worker/NLog.xsd
```diff
@@ -0,0 +1,3103 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
elliotritchie: Is this file needed?

DScuffle: Nope - Removing

src/SFA.DAS.EAS.MessageHandlers.Worker/NLog.config
DScuffle: Nope - Removing

src/SFA.DAS.EAS.MessageHandlers.Worker/EventHandlers/ApprovedTransferConnectionInvitationEventHandler.cs
DScuffle: Agreed - extracted.

PR #1216, opened 2018-04-24
AML-2252 Update About Your Agreement Copy

DScuffle:
Copy only update - Needs to go when we have a transfers agreement available


elliotritchie:
Merged with PR #1214.

PR #1223, opened 2018-04-30
update-packages

gdbarnes:


gdbarnes:
I followed steps from here to fix this issue:
https://github.com/Microsoft/vscode/issues/48783

Looks like a recent issue where a load of npm packages need to update their 'hoek' dependancy to a later version.

PR #1230, opened 2018-05-02
Adds service bus subscription for Das-Data transfer relationships

desderry:
Allows Das-data function to process transfer relationship status messages


elliotritchie:
Cherry picked into PR #1225.

PR #1265, opened 2018-06-07
AML-2386 Removed message that is no longer being used

robinoconnor:


elliotritchie:
Cherry picked into PR #1251.

PR #1288, opened 2018-07-16
AML-2502 Added payment filter to remove all fully funded payments from new pay…

robinoconnor:
…ments to be saved to database


src/SFA.DAS.EAS.Employer_Financial.Database/Scripts/Manual/AML-2502_DeleteFullyFundedPayments.sql
```diff
@@ -0,0 +1,6 @@
+﻿DELETE meta FROM [employer_financial].[PaymentMetaData] meta
+INNER JOIN [employer_financial].[Payment] p ON p.PaymentMetaDataId = meta.Id
+WHERE p.FundingSource = 4
+
+DELETE FROM [employer_financial].[Payment]
```
elliotritchie: I think payments will need to be deleted before meta data since the foreign key is on the Payments table.

PR #1291, opened 2018-07-18
AML-2534 - Update nuget packages

iangledhill:
This updates activities, nlog and das.nlog to remove a problem introduced into das.nlog when a ctor was changed


iangledhill:
This branch was merged into AML-2116

PR #1293, opened 2018-07-20
Aml 2360 rework acceptance test n service bus

DScuffle:
This is the foundation work for the acceptance test, it sets up 2 running add/remove PayeScheme tests, which call into the controller to initiate the action being tested.


elliotritchie:
Looks like this is an obsolete branch with `AML-2360-Rework-HMRC-AcceptanceTests-New-Projects` having newer commits.

PR #1299, opened 2018-07-26
AML-2445 Add authentication to employer accounts & employer finance

elliotritchie:


elliotritchie:
Cherry picked into #1280

PR #1301, opened 2018-08-02
Update activities client to fix missing activities

iangledhill:
Updated activities client nuget packages


iangledhill:
Have merged this branch into AML-2534, as they are both heading towards the same goal

PR #1315, opened 2018-08-10
aml-2660-Fix-Register-For-Invite-Logged-In-Issue

DScuffle:
Adds a SignOut attribute, applies it to the register and accept action, to make sure we don't register a new user whilst logged in!


elliotritchie:
Cherry picked into #1335.

PR #1328, opened 2018-08-15
Aml 2119 updated journey

iangledhill:
This changes the user journey when updating an organisation's details.

I've also changed the Organisation Type aka Source to be (almost) uniformly the common OrganisationType enum. Previously it was defined as variously as 8 and 16 bit ints. Not possibly to change it in the types assembly (so the LegalEntityViewModel) as I didn't want to introduce a dependency on the common types into the project.


gdbarnes:
Styling of new and existing views now complete.

PR #1343, opened 2018-08-24
AML-2691 Filtered out levy with no YTD values as they have caused issues in prod

robinoconnor:
Pritesh want's us to filter out Levy with no YTD values for now to fix the issues we are having on prod with declarations not being processed due to end of year adjustments having YTD values set to null. The code throws an error if this occurs to i've filtered them out and created a warning instead that gets logged


elliotritchie:
Closed in favour of PR #1360.

PR #1356, opened 2018-09-06
Aml 2457 create employer accounts and finance api projects

iangledhill:
These are just bare bones projects


src/SFA.DAS.EmployerFinance.Api/packages.config
```diff
@@ -0,0 +1,10 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
elliotritchie: Can this be removed?

PR #1367, opened 2018-09-14
Aml 2840 fix query performance

iangledhill:
Retired stored proc and replaced with EF

### DIFF
--- a/src/SFA.DAS.EmployerApprenticeshipsService.Infrastructure/Data/EmployerAccountRepository.cs
+++ b/src/SFA.DAS.EmployerApprenticeshipsService.Infrastructure/Data/EmployerAccountRepository.cs
@@ -10,6 +10,8 @@ using System.Data;
 using System.Data.Entity;
 using System.Linq;
 using System.Threading.Tasks;
+using SFA.DAS.Authorization;
+using SFA.DAS.EAS.Domain.Models.EmployerAgreement;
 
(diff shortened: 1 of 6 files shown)


src/SFA.DAS.EmployerApprenticeshipsService.Infrastructure/Data/EmployerAccountRepository.cs
```diff
@@ -69,37 +71,36 @@ public async Task<Accounts<Domain.Models.Account.Account>> GetAccounts(string to
...
-
-            await _db.Value.Database.Connection.QueryAsync<AccountDetail, string, long, AccountDetail>(
-                sql: "[employer_account].[GetAccountDetails_ByHashedId]",
-                param: parameters,
-                transaction: _db.Value.Database.CurrentTransaction.UnderlyingTransaction,
-                commandType: CommandType.StoredProcedure,
-                splitOn: "PayeSchemeId,LegalEntityId",
-                map: (parent, payeSchemeRef, legalEntityId) =>
+            var accountDetail = await _db.Value.Accounts
+                .Where(ac => ac.HashedId == hashedAccountId)
```
elliotritchie: I know this is how it worked before but we shouldn't be looking up accounts using the hashed ID. I'm happy for this repository method to be deleted and for its body to be moved into the query handler itself if that helps.

PR #1370, opened 2018-09-18
Add correct connection reference

robinoconnor:


src/SFA.DAS.EmployerApprenticeshipsService.Infrastructure/Data/EmployerAccountRepository.cs
```diff
@@ -55,10 +55,9 @@ public async Task<Accounts<Domain.Models.Account.Account>> GetAccounts(string to
                 sql: $"select count(*) from [employer_account].[Account] a;",
                 transaction: _db.Value.Database.CurrentTransaction.UnderlyingTransaction);
 
-            var result = await WithConnection(async c => await c.QueryAsync<Domain.Models.Account.Account>(
+            var result = await _db.Value.Database.Connection.QueryAsync<Domain.Models.Account.Account>(
                 sql: $"select a.* from [employer_account].[Account] a ORDER BY a.Id OFFSET {offset} ROWS FETCH NEXT {pageSize} ROWS ONLY;",
-                transaction: _db.Value.Database.CurrentTransaction.UnderlyingTransaction,
-                commandType: CommandType.Text));
+                transaction: _db.Value.Database.CurrentTransaction.UnderlyingTransaction);
```
iangledhill: The dbConnection is already enlisted in the transaction by this point. If a txn object is supplied here all that happens is that it is checked to make sure it is the same txn as the one the connection is already enlisted in. Therefore L60 (setting the txn) is redundant.

PR #1380, opened 2018-09-25
Add logging to record invalid hashed ids

iangledhill:
Added logging to record invalid/unknown hashed ids and also to log successful reads

### DIFF
--- a/src/SFA.DAS.EmployerApprenticeshipsService.Infrastructure/Data/EmployerAccountRepository.cs
+++ b/src/SFA.DAS.EmployerApprenticeshipsService.Infrastructure/Data/EmployerAccountRepository.cs
@@ -8,21 +8,25 @@ using System;
 using System.Collections.Generic;
 using System.Data;
 using System.Data.Entity;
+using System.Diagnostics;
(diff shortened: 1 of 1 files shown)


src/SFA.DAS.EmployerApprenticeshipsService.Infrastructure/Data/EmployerAccountRepository.cs
```diff
@@ -8,21 +8,25 @@
...
     public class EmployerAccountRepository : BaseRepository, IEmployerAccountRepository
     {
         private readonly Lazy<EmployerAccountsDbContext> _db;
+        private readonly ILog _logger;
 
         public EmployerAccountRepository(EmployerApprenticeshipsServiceConfiguration configuration, ILog logger, Lazy<EmployerAccountsDbContext> db)
             : base(configuration.DatabaseConnectionString, logger)
         {
             _db = db;
+            _logger = logger;
```
DScuffle: This is already done in base class, it doesn't need to be done here too (but no problem if it remains as it is more explicit doing it here)

PR #1390, opened 2018-10-02
Added data fix to pre-deployment folder

PeterDavidWilliams:
No code changes as the query already pulls from the database so just a datafix to fix any NULL entries in PaymentMetaData for ApprenticeshipCourseName and PathwayName.

### DIFF
--- a/src/SFA.DAS.EAS.Employer_Financial.Database/Scripts/PreDeployment/AML-2671-PopulateNullApprenticeshipCourseNames.sql
+++ b/src/SFA.DAS.EAS.Employer_Financial.Database/Scripts/PreDeployment/AML-2671-PopulateNullApprenticeshipCourseNames.sql
@@ -19,7 +19,7 @@ FROM (
     WHERE    PathwayName IS NOT NULL
             AND    ApprenticeshipCourseName IS NOT NULL) AS T1
 WHERE T1.RowNumber = 1
-ORDER BY 1, 2, 3
+ORDER BY 1, 2, 3;
 
 		UPDATE Bad
 		SET Bad.ApprenticeshipCourseName = FW.ApprenticeshipCourseName,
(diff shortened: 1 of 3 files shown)


src/SFA.DAS.EAS.Employer_Financial.Database/Scripts/PreDeployment/AML-2671-PopulateNullApprenticeshipCourseNames.sql
```diff
@@ -0,0 +1,50 @@
+﻿BEGIN TRY
+BEGIN TRANSACTION
```
iangledhill: It doesn't make much of a difference to the execution of the SQL but the BEGIN TRAN is within the BEGIN TRY but the COMMIT TRAN is outside, which makes the scope of the TRY and TRAN overlap slightly. It's usually better to have purely nested scopes (this is generally true in code - cognitively it makes it easier to deal with). To achieve this L49-50 should be moved to L34 (and we don't need L49 anymore, as tran count will always be 1 at that point).

src/SFA.DAS.EAS.Employer_Financial.Database/Scripts/PreDeployment/AML-2671-PopulateNullApprenticeshipCourseNames.sql
```diff
@@ -0,0 +1,50 @@
...
+
+DECLARE @Frameworks as table(
+        FrameworkCode int,
+        ProgrammeType int,
+        PathwayCode int,
+        PathwayName varchar(max),
+        ApprenticeshipCourseName varchar(max));
+
+INSERT INTO @Frameworks (FrameworkCode, ProgrammeType, PathwayCode, PathwayName, ApprenticeshipCourseName)
```
iangledhill: All statements _should_ end with a semi-colon - in a future version it will be mandatory.
https://docs.microsoft.com/en-us/sql/t-sql/language-elements/transact-sql-syntax-conventions-transact-sql?view=sql-server-2017

src/SFA.DAS.EAS.Employer_Financial.Database/Scripts/PreDeployment/AML-2671-PopulateNullApprenticeshipCourseNames.sql
PeterDavidWilliams: Suggested changes made.

src/SFA.DAS.EAS.Employer_Financial.Database/Scripts/PreDeployment/AML-2671-PopulateNullApprenticeshipCourseNames.sql
PeterDavidWilliams: Suggested changes made.

PR #1402, opened 2018-10-09
Aml 2698 fix handling of multiple adjustments in the same year

iangledhill:
This separates out the LD handling out from the command handler to allow the processing to be more easily tested.
There are UTs around the new class to cover the behaviour described in this ticket and also for ticket AML-2855.


src/SFA.DAS.EmployerFinance/Commands/RefreshEmployerLevyData/LevyImportCleanupStrategy.cs
```diff
@@ -0,0 +1,191 @@
...
+        /// <returns>
+        ///     The declaration that was effective for period 12 or null if one was not found.
+        /// </returns>
+        /// <remarks>
+        ///     The database contains all the declarations retrieved to date from HMRC, <see cref="hmrcDeclarations"/>
+        ///     contains the declarations that HMRC has that we have not retrieved previously (which is determined by submission date).
+        /// </remarks>
+        private async Task<DasDeclaration> GetDeclarationEffectiveForPeriod12(string empRef, string payrollYear, DateTime yearEndAdjustmentCutOff, IEnumerable<DasDeclaration> hmrcDeclarations)
+        {
+            // Look in the declarations that have just been retrieved from HMRC (which will contain everything not yet in the database)
```
MichaelYoung1981: This comment is unnecessary, the method below is very well named so is descriptive enough.

src/SFA.DAS.EmployerFinance/Commands/RefreshEmployerLevyData/LevyImportCleanupStrategy.cs
```diff
@@ -0,0 +1,191 @@
...
+            ILog logger)
+        {
+            _dasLevyRepository = dasLevyRepository;
+            _hmrcDateService = hmrcDateService;
+            _logger = logger;
+        }
+
+        public async Task<IEnumerable<DasDeclaration>> Cleanup(string empRef, IEnumerable<DasDeclaration> declarations)
+        {
+            var temp = declarations.OrderBy(c => c.SubmissionDate).ToArray();
```
MichaelYoung1981: Rename temp to something I won't forget in 4 lines. :)

src/SFA.DAS.EmployerFinance/Commands/RefreshEmployerLevyData/LevyImportCleanupStrategy.cs
```diff
@@ -0,0 +1,191 @@
...
+using System.Threading.Tasks;
+using SFA.DAS.EmployerFinance.Data;
+using SFA.DAS.EmployerFinance.Extensions;
+using SFA.DAS.EmployerFinance.Models.Levy;
+using SFA.DAS.EmployerFinance.Services;
+using SFA.DAS.NLog.Logger;
+
+namespace SFA.DAS.EmployerFinance.Commands.RefreshEmployerLevyData
+{
+    public class LevyImportCleanerStrategy : ILevyImportCleanerStrategy
```
MichaelYoung1981: I'm not sure this is actually an implementation of the strategy pattern.  I can't imagine there will ever be more than one?

src/SFA.DAS.EmployerFinance/Commands/RefreshEmployerLevyData/ILevyImportCleanupStrategy.cs
```diff
@@ -0,0 +1,19 @@
...
+    ///     Represents a service that can take levy declarations and pre-process them to make them 
+    ///     suitable for processing by EAS. 
+    /// </summary>
+    /// <remarks>
+    ///     The pre-processing that occurs is to remove duplicate subsidy ids, remove declarations that 
+    ///     are outside the period range and evaluate year end adjustments. 
+    /// </remarks>
+    public interface ILevyImportCleanerStrategy
+    {
+        Task<IEnumerable<DasDeclaration>> Cleanup(string empRef, IEnumerable<DasDeclaration> declarations);
```
MichaelYoung1981: As the interface is really internal I'm not convinced we need to be so open on the input and output types given its forcing us to do a ToArray on both sides.
I'm not actually convinced we need an interface but that's a bigger discussion.

src/SFA.DAS.EmployerFinance/Commands/RefreshEmployerLevyData/LevyImportCleanupStrategy.cs
```diff
@@ -0,0 +1,191 @@
...
+                yearEndAdjustment.EndOfYearAdjustmentAmount =
+                    period12Declaration.LevyDueYtd.Value - yearEndAdjustment.LevyDueYtd ?? 0;
+            }
+            else
+            {
+                yearEndAdjustment.EndOfYearAdjustmentAmount = -yearEndAdjustment.LevyDueYtd ?? 0;
+            }
+        }
+
+        /// <summary>
```
MichaelYoung1981: I'm not convinced on the need to include private method summary comments.  The majority of what is written should be obvious to the person reading and the methods and parameters are well enough named to give clarity.

src/SFA.DAS.EmployerFinance/Commands/RefreshEmployerLevyData/RefreshEmployerLevyDataCommandHandler.cs
```diff
@@ -55,22 +54,16 @@ protected override async Task HandleCore(RefreshEmployerLevyDataCommand message)
 
             foreach (var employerLevyData in message.EmployerLevyData)
             {
-                var declarations = employerLevyData.Declarations.Declarations.OrderBy(c => c.SubmissionDate).ToArray();
+                var declarations = await _levyImportCleanerStrategy.Cleanup(employerLevyData.EmpRef, employerLevyData.Declarations.Declarations);
 
-                declarations = FilterDuplicateHmrcDeclarations(employerLevyData.EmpRef, declarations);
+                var temp = declarations.ToArray();
```
MichaelYoung1981: Rename temp or change the return type of Cleanup.

src/SFA.DAS.EmployerFinance/Commands/RefreshEmployerLevyData/LevyImportCleanupStrategy.cs
```diff
@@ -0,0 +1,191 @@
...
+            return period12Declaration;
+        }
+
+        private Task<DasDeclaration> GetEffectivePeriod12SubmissionFromLatestHmrcFeed(IEnumerable<DasDeclaration> declarations, string payrollYear, DateTime yearEndAdjustmentCutOff)
+        {
+            // We are only interested in declarations for the current year that are not year-end-adjustments and that were submitted on time
+            DasDeclaration period12Declaration = declarations
+                .Where(ld => ld.PayrollYear == payrollYear
+                             && ld.PayrollMonth.HasValue
+                             // We're interested in LD that are either within the month (i.e. pre-cut off) or are year end adjustments before the supplied cut-off date
```
MichaelYoung1981: Extract the condition into a method and remove the comment?

src/SFA.DAS.EmployerFinance/Commands/RefreshEmployerLevyData/LevyImportCleanupStrategy.cs
```diff
@@ -0,0 +1,191 @@
...
+            {
+                period12Declaration = await GetEffectivePeriod12SubmissionFromDatabase(empRef, payrollYear, yearEndAdjustmentCutOff);
+            }
+
+            return period12Declaration;
+        }
+
+        private Task<DasDeclaration> GetEffectivePeriod12SubmissionFromLatestHmrcFeed(IEnumerable<DasDeclaration> declarations, string payrollYear, DateTime yearEndAdjustmentCutOff)
+        {
+            // We are only interested in declarations for the current year that are not year-end-adjustments and that were submitted on time
```
MichaelYoung1981: The method name could probably be tweaked to remove the comment although it makes sense to me.

src/SFA.DAS.EmployerFinance.UnitTests/Commands/RefreshPaymentDataTests/LevyImportCleanupStrategyTests.cs
```diff
@@ -0,0 +1,245 @@
...
+        public LevyImportCleanerStrategy CreateStrategy()
+        {
+            return new LevyImportCleanerStrategy(
+                DasLevyRepository,
+                HmrcDateService,
+                Log
+                );
+        }
+
+        public LevyImportCleanerStrategyTestFixtures WithDeclarations(params long[] subsidyIds)
```
MichaelYoung1981: What's a subsidy id?

src/SFA.DAS.EAS.Employer_Financial.Database/StoredProcedures/GetEffectivePeriod12Declaration.sql
```diff
@@ -0,0 +1,32 @@
...
+		DateCeased,
+		InactiveFrom,
+		InactiveTo,
+		HmrcSubmissionId,
+		NoPaymentForPeriod
+	FROM 
+		[employer_financial].[GetLevyDeclaration]
+	WHERE	EmpRef = @EmpRef
+			AND	PayrollYear = @PayrollYear
+			AND (LastSubmission = 1 OR (EndOfYearAdjustment = 1 AND SubmissionDate < @yearEndAdjustmentCutOff))
```
MichaelYoung1981: I'm being lazy...do we need to make sure the table has an appropriate index?

src/SFA.DAS.EmployerFinance/Commands/RefreshEmployerLevyData/LevyImportCleanupStrategy.cs
iangledhill: Removed

src/SFA.DAS.EmployerFinance/Commands/RefreshEmployerLevyData/LevyImportCleanupStrategy.cs
iangledhill: Renamed

src/SFA.DAS.EmployerFinance/Commands/RefreshEmployerLevyData/LevyImportCleanupStrategy.cs
iangledhill: Agreed it's definitely a small family, but the strategy pattern only describes injecting behaviour from a family of behaviours but does not specify a minimum size of the family.

src/SFA.DAS.EmployerFinance/Commands/RefreshEmployerLevyData/ILevyImportCleanupStrategy.cs
iangledhill: Unfortunately the interface is injected into a public constructor (so must be public itself) and if it's public it should play nicely. 

I'm not sure that changing the input parameter to an array would not remove the need to call ToArray(); the inbound declarations are ordered by submission date (creating another enumeration) which would _still_ need to be projected to an array.

I've changed the return type to an array.

src/SFA.DAS.EmployerFinance/Commands/RefreshEmployerLevyData/RefreshEmployerLevyDataCommandHandler.cs
iangledhill: I've changed the return type.

src/SFA.DAS.EmployerFinance/Commands/RefreshEmployerLevyData/LevyImportCleanupStrategy.cs
iangledhill: Done

src/SFA.DAS.EmployerFinance.UnitTests/Commands/RefreshPaymentDataTests/LevyImportCleanupStrategyTests.cs
iangledhill: Fixed

src/SFA.DAS.EAS.Employer_Financial.Database/StoredProcedures/GetEffectivePeriod12Declaration.sql
iangledhill: The query is narrower than the sproc used previously by the repo but I've checked and there is a covering index on empref and payroll year. Given that an empref would be expected to typically make 12 declarations per year that should be fine.

PR #1409, opened 2018-10-10
Aml 2395 drop deprecated view

iangledhill:
Drop view employer_financial.GetLevyDeclarations


src/SFA.DAS.EAS.Employer_Financial.Database/Scripts/PreDeployment/AML-2395-DropDeprecatedView.sql
```diff
@@ -0,0 +1,7 @@
+﻿
+IF OBJECT_ID('[employer_financial].[GetLevyDeclarations]') > 0
```
DScuffle: Should this be checking explicitly for a view i.e. sys.views

PR #1412, opened 2018-10-11
Aml 2904 hmrc retry policy

robinoconnor:


src/SFA.DAS.UnitTests/ExecutionPolicies/HmrcExecutionPolicyTests/PolicyRetryTest.cs
```diff
@@ -0,0 +1,94 @@
...
+            public int MaxCallCount { get; }
+
+            public TestRunner(T exception, int maxCallCount = 10)
+            {
+                _exception = exception;
+                MaxCallCount = maxCallCount;
+                CallCount = 0;
+            }
+
+            public async Task Execute()
```
iangledhill: The method shouldn't be async (since the method is not awaiting the results of a task) and the await at line 88 is artificial just top keep the compiler happy. The slightly odd-looking line 88 can go. The method would return a task so line 86 would then become return Task,Completed. 
Does L83 look a bit lost by itself, do you thiink? L83 and L85 could be combined so the post-increment operator is part of the expression (it would need to be pre-increment operator to maintain the same logic).

src/SFA.DAS.UnitTests/ExecutionPolicies/HmrcExecutionPolicyTests/PolicyRetryTest.cs
```diff
@@ -0,0 +1,94 @@
...
+        {
+            _policy = new HmrcExecutionPolicy(Mock.Of<ILog>(), TimeSpan.FromMilliseconds(0));
+        }
+
+        [Test]
+        public void LimitedRetryWhenServiceUnavailableReturned()
+        {
+            var runner = new TestRunner<ApiHttpException>(CreateTestException(503, "service unavailable"));
+
+            Assert.Throws<AggregateException>(() => _policy.ExecuteAsync(runner.Execute).Wait());
```
iangledhill: Is it worth adding a timeout to the wait()? Since the behaviour of the retry logic is being tested it is possible that it might not work and just keep re-trying for ever, which would cause the test to hang. Adding a timeout would guard against this. 
I'm always suspicious of blocking calls. Could these be avoided by returning a task from the test and/or using throwsasync() may be?

src/SFA.DAS.UnitTests/ExecutionPolicies/HmrcExecutionPolicyTests/PolicyRetryTest.cs
robinoconnor: changes made

src/SFA.DAS.UnitTests/ExecutionPolicies/HmrcExecutionPolicyTests/PolicyRetryTest.cs
robinoconnor: changes made

robinoconnor:
This has passed review but we can't change status atm

PR #1444, opened 2018-10-24
AML-2460 - Move PAYE pages to the accounts site

iangledhill:
This requires new config from the AML-2460 branch in employer-config repo


src/SFA.DAS.EmployerAccounts/Attributes/ServiceBusConnectionKeyAttribute.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+
+namespace SFA.DAS.EmployerAccounts.Attributes
+{
+    public class ServiceBusConnectionKeyAttribute : Attribute
+    {
+        public ServiceBusConnectionKeyAttribute(string connectionKey)
```
robinoconnor: It looks like the connectionKey on this is not used anywhere. I'm aware this may have just been copied over but is it worth just removing it now? unless it will be used later on with code that hasn't been moved yet?

iangledhill: This constructor is in the existing code base and is called. The PRs are already extremely large but small-scale, localised clean-up might be okay if significant (i.e. a warning or dangerous code) but changing a public signature and fixing up consequent references is probably left for another ticket to avoid PR bloat.

robinoconnor: ok thats fine, we need to review the code after the decoupling anyway.

PR #1452, opened 2018-10-30
Fix problem with get transactions always returning atleast one line

iangledhill:
The purpose of this change is to prevent the null UKPRNs being passed on the transaction screen.
This looks like it is being caused because the SP [employer_financial].[GetTransactionLines_ByAccountId] will always return a transaction, even when there aren't any (because it left joins on to the account balance, which will always exist). It looks like in the past the code has expected the SP to return zero rows because there is branching logic to handle no rows (which will currently never be exercised). The SQL and code have maybe diverged over time. 
So this fix removes the balance from the stored proc result set (TransactionLine) and moves it up to the same level as the collection of transaction lines (AggregationData) so the balance is always available, whether there are any transactions or not. 
The mediator handlers invoked by the IDasService have been removed and the repo calls these contained have been elevated up to the IDasService implementation. The IDasService could also be removed, but decided against this. 
The IDasService return types have been changed from List<T> and ICollection<T> to straight arrays<T>, since this is a more appropriate type and is more efficient.

### DIFF
--- a/src/SFA.DAS.EAS.Web/Views/Shared/_Layout.cshtml
+++ b/src/SFA.DAS.EAS.Web/Views/Shared/_Layout.cshtml
@@ -10,6 +10,19 @@
 <html lang="en">
 <!--<![endif]-->
 <head>
+
+    <!-- Google Tag Manager -->
+    <script>
+    (function (w, d, s, l, i) {
+        w[l] = w[l] || []; w[l].push({
+            'gtm.start':
+                new Date().getTime(), event: 'gtm.js'
+        }); var f = d.getElementsByTagName(s)[0],
+            j = d.createElement(s), dl = l != 'dataLayer' ? '&l=' + l : ''; j.async = true; j.src =
+                'https://www.googletagmanager.com/gtm.js?id=' + i + dl; f.parentNode.insertBefore(j, f);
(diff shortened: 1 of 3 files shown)


src/SFA.DAS.EmployerFinance/Services/DasLevyService.cs
```diff
@@ -1,99 +1,104 @@
...
-                FromDate = fromDate,
-                ToDate = toDate
-            });
+            return _transactionRepository.GetAccountBalance(accountId);
+        }
 
-            return result.TransactionLines;
+        public Task<TransactionLine[]> GetAccountTransactionsByDateRange(long accountId, DateTime fromDate, DateTime toDate)
+        {
+            return _transactionRepository.GetAccountTransactionsByDateRange(accountId, fromDate, toDate);
```
robinoconnor: This method gets called by EmployerAccountTransactionsOrchestrator which uses the PaymentDate L377.
Also is this paymentDate something that is could be used if someone directly called the repository as it may need to be moved if this is the case to make sure its always set even if you do not get transactionLines without the use of the DasService

robinoconnor: After discussing with Ian it was decided that this change should be investigated as a separate tech task to see if it is a bug that needs fixing or whether the code is meant to work in that way as the original code did not have the check.

PR #1457, opened 2018-11-02
AML-2998 levy analyzer

iangledhill:
This is a CLI that will run rules against levy and transactions. This only contains a couple of rules (the others will come in a separate PR) but does contain a working CLI.


DannyGB:
The original author has left and the PR is over 12 months old and will never be merged into master

PR #1459, opened 2018-11-05
AML-2989 Consolidate UserAccountSettings Table

DScuffle:
This PR moves the ReceivesNotification setting to the membership table and removes the user account settings table. All references to the UAS have been updated. I have manually tested the removal and setting of RecievesNotification and I have initiated a transfer which checks this setting to see if an email should be sent. 

By consolidating the two tables we will remove a lot of PROD bugs that arise where we get a record in one of these tables and not the other (which our system does not handle).


DannyGB:
This will never be merged after 13 months

PR #1476, opened 2018-11-22
AML-2645 incorrect message when remove org

scottywakefield:
There's a lot of 'noise' in this PR due to updating the nugets required, in commit `ceb314b`.


src/SFA.DAS.EAS.Employer_Account.Database/StoredProcedures/GetEmployerAgreementsToRemove_ByAccountId.sql
```diff
@@ -9,7 +9,9 @@ BEGIN
 			ALE.Name, 
 			ISNULL(Signed.StatusId, Pending.StatusId) AS Status, 
 			a.HashedId, 
-			le.Code
+			le.Code as LegalEntityCode,
+			le.[Source] as LegalEntitySource,
+			ale.PublicHashedId as HashedAgreementId
```
robinoconnor: That PublicHashedId  on ALE is not the agreement hashed id but the legal entity hashed id so this needs to be changed. I'm not even sure we have an agreement hashed ID in the database so you might need to unhash it and then compare against the id itself.

src/SFA.DAS.EmployerApprenticeshipsService.Domain/Models/EmployerAgreement/RemoveEmployerAgreementView.cs
```diff
@@ -9,5 +11,6 @@ public class RemoveEmployerAgreementView
         public string HashedAccountId { get; set; }
         public bool CanBeRemoved { get; set; }
         public string LegalEntityCode { get; set; }
+        public OrganisationType LegalEntitySource { get; set; }
```
robinoconnor: I'd call this LegalEntityOrganisationType as it will confuse other devs if they see a source being compared to a organisation type.

scottywakefield: I called it that to match what is on `SFA.DAS.EAS.Domain.Models.EmployerAgreement.EmployerAgreementView`. to keep them consistent. Do you want me to change that one as well? Not sure if there would be side effects.

src/SFA.DAS.EAS.Employer_Account.Database/StoredProcedures/GetEmployerAgreementsToRemove_ByAccountId.sql
scottywakefield: in `GetEmployerAgreement.sql` it has the following:
`ale.PublicHashedId as AccountLegalEntityPublicHashedId`, which is the same as HashedAgreementId right?

robinoconnor: no, it's the hashed id of the legal entity that is associated with the agreement. As far as I can tell it is never used as the hashed agreement id

src/SFA.DAS.EmployerApprenticeshipsService.Application.UnitTests/Commands/RemoveLegalEntityTests/WhenIValidateTheCommand.cs
```diff
@@ -99,18 +100,67 @@ public async Task ThenTheAgreementIsCheckedToSeeIfItHasBeenSignedAndHasActiveCom
...
+
+            //Act
+            var actual = await _removeLegalEntityCommandValidator.ValidateAsync(new RemoveLegalEntityCommand
+            {
+                HashedAccountId = ExpectedHashedAccountId,  UserId = ExpectedUserId, HashedLegalAgreementId = ExpectedHashedAgreementId
+            });
+
+            //Assert
+            Assert.IsTrue(actual.IsValid());
+            //Assert.Contains(new KeyValuePair<string, string>("HashedLegalAgreementId", "Agreement has already been signed and has active commitments"), actual.ValidationDictionary);
```
robinoconnor: does this need to be removed as its commented out?

scottywakefield: 😮 oops!

PR #1478, opened 2018-11-23
Aml 2469 2464 2465 2468 2471 2466 2470 recreate apis

DScuffle:
This migrates All the API's and has a clean-up.

I have tested every api as a url and through the legacy client. The behaviour is consistent (there is one api broken in legacy EmployerAgreements, that I have fixed in the new clients, but I have chosen to retain this behaviour as we want to be consistent and it complicates the release fixing it now)


src/SFA.DAS.EmployerFinance/Models/Levy/LevyDeclarationView.cs
```diff
@@ -0,0 +1,57 @@
...
+            {
+                return null;
+            }
+
+            var year = 2000;
+            var month = 1;
+
+            if (PayrollMonth <= 9)
+            {
+                year += Convert.ToInt32(PayrollYear.Split('-')[0]);
```
robinoconnor: @robinoconnor check

src/SFA.DAS.EmployerFinance/Queries/GetStatistics/GetStatisticsQueryHandler.cs
```diff
@@ -0,0 +1,39 @@
...
+
+        public GetStatisticsQueryHandler(EmployerFinanceDbContextReadOnly financeDb)
+        {
+            _financeDb = financeDb;
+        }
+
+        public async Task<GetStatisticsResponse> Handle(GetStatisticsQuery message)
+        {
+            // TODO When decoupling
+            //var accountsQuery = _accountDb.Value.Accounts.FutureCount();
```
robinoconnor: Should these still be uncommented?

src/SFA.DAS.EAS.Api/Extensions/UrlHelperExtensions.cs
```diff
@@ -0,0 +1,38 @@
...
+
+            return Action(baseUrl, pathAndQuery);
+        }
+
+        private static string Action(string baseUrl, string path)
+        {
+            var trimmedBaseUrl = baseUrl?.TrimEnd('/') ?? string.Empty;
+            var trimmedPath = path?.TrimStart('/') ?? string.Empty;
+
+            return $"{trimmedBaseUrl}/{trimmedPath}".TrimEnd('/');
```
robinoconnor: You can just do a Trim rather than a TrimStart above and then you don't need the TrimEnd here. You also might want to have a whitespace char to the trim params as well to remove any empty space.

src/SFA.DAS.EmployerFinance.UnitTests/Queries/GetStatisticsTests/GetStatisticsQueryTests.cs
```diff
@@ -0,0 +1,107 @@
...
+    public class GetStatisticsQueryTests : FluentTest<GetStatisticsQueryTestsFixtures>
+    {
+        [Test]
+        public Task Handle_WhenIGetStatistics_ThenShouldReturnResponse()
+        {
+            return RunAsync(f => f.Handle(), (f, r) => r.Should().NotBeNull()
+                .And.Match<GetStatisticsResponse>(r2 =>
+                    r2.Statistics != null &&
+                    // Todo: replace when decoupled
+                    //r2.Statistics.TotalAccounts == 2 &&
```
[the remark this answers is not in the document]
DScuffle: Removed

src/SFA.DAS.EmployerFinance/Queries/GetStatistics/GetStatisticsQueryHandler.cs
DScuffle: Removed

src/SFA.DAS.EAS.Api/Extensions/UrlHelperExtensions.cs
DScuffle: Taken down to one Trim

DannyGB:
This will never be merged after 14 months

PR #1482, opened 2018-11-27
Link added to ma homepage for provider permissions

tom-gough:


src/SFA.DAS.EmployerAccounts/Configuration/EmployerAccountsConfiguration.cs
```diff
@@ -14,6 +14,7 @@ public class EmployerAccountsConfiguration : ITopicMessagePublisherConfiguration
         public string EmployerPortalBaseUrl { get; set; }
         public string EmployerProjectionsBaseUrl { get; set; }
         public string EmployerRecruitBaseUrl { get; set; }
+        public string ProviderPermissionsBaseUrl { get; set; }
```
elliotritchie: Can this be renamed to `ProviderRelationships` to match the application/feature name? This will also need adding to the `SFA.DAS.EmployerAccounts.json`/`SFA.DAS.EmployerAccounts.schema.json` config (I recommend using `https://localhost:44379` as the default) https://github.com/SkillsFundingAgency/das-employer-config/blob/master/das-employerapprenticeshipsservice

PR #1485, opened 2018-11-28
AML-3061 fix integration tests

scottywakefield:
Issues still unresolved:

1. Split out the data setup code more from the api hosting code.
1. Use a different IOC container for db setup and running the api (we can't test IOC is working if we're freaking with it). the point above will facilitate this.
1. General tidy up. The tests and API now both work but a general tidy up would be well worth it.


src/SFA.DAS.EAS.Web.IntegrationTests/TestUtils/DataHelper/DbBuilder.cs
```diff
@@ -13,21 +15,38 @@ class DbBuilder
...
+        {
+            _dbContext.Database.BeginTransaction();
+            return this;
+        }
+
+        public DbBuilder CommitTransaction()
+        {
+            if (_dbContext.Database.CurrentTransaction == null) return this;
+
+            _dbContext.Database.CurrentTransaction.Commit();
```
iangledhill: I wondered whether it would be worthwhile guarding against errant calls? e.g. multiple calls to BeginTransaction and CommitingTransaction? Pretty sure errant calls will be faulted anyway by MSSQL but instinctively feel the class should be defending/handling this itself. However, this is a test helper class and simplicity > sophistication so perhaps as it is might be better.

PR #1491, opened 2018-11-30
Integration tests: statistics

scottywakefield:
Summary of additional changes:
1. removed package-lock.json files from repo
1. added and using fluent assertions
1. rearranged db builder folders and added new data helpers


scottywakefield:
@robinoconnor I thought about trying to add a pre-known amount but that would require first removing any existing records and then adding in the new ones. For each of the totals. So instead I just read out whatever is the currently expected amounts (which could all be zero, but usually not) and checking against what is returned in the actuals (which could all be zeros if not mapped correctly). So yes, I agree that it's not perfect but it does test more than it used to. And occasionally (when connected to a db with more than 0 rows) it will test it quite a bit. I hope that answers the question.

src/SFA.DAS.EAS.Web.IntegrationTests/HealthCheckControllerTests/WhenHealthCheck.cs
```diff
@@ -17,10 +17,10 @@ public async Task ThenTheStatusShouldBeOk()
                             .AllowStatusCodes(HttpStatusCode.OK);
 
             // Act
-            await ApiIntegrationTester.InvokeIsolatedGetAsync(call);
+            var response = await ApiIntegrationTester.InvokeIsolatedGetAsync(call);
 
             // Assert
-            Assert.Pass("Verified we got OK");
+            Assert.That(response.Response.StatusCode == HttpStatusCode.OK);
```
iangledhill: This is minor but I think this assertion only tests that it is true. Using Assert.Equals(…) or the fluent equivalent will show the expected value and the actual value , which would be more useful in the test messages.

src/SFA.DAS.EAS.Web.IntegrationTests/TestUtils/DataAccess/DataHelpers/AccountStatisticsDataHelper.cs
```diff
@@ -0,0 +1,97 @@
...
+                .Build<User>()
+                .Without(user => user.Id)
+                .Without(user => user.UserRef)
+                .Create();
+            var accountRepo = new AccountRepository(_configuration,
+                Mock.Of<ILog>(), lazyDb, Mock.Of<IAccountLegalEntityPublicHashingService>());
+
+            try
+            {
+                accountDbContext.Database.BeginTransaction();
```
iangledhill: Should the begin txn be outside the try? Currently if the begin txn faults a txn rollback will be attempted, which will fault. Another occurrence below.

PR #1500, opened 2018-12-05
GetLegalEntity

PeterDavidWilliams:


src/SFA.DAS.EAS.Web.IntegrationTests/LegalEntitiesControllerTests/WhenIGetLegalEntitiesWithKnownIds.cs
```diff
@@ -0,0 +1,59 @@
...
+            var callRequirements = new CallRequirements($"api/accounts/{hashedAccountId}/legalentities")
+                .ExpectControllerType(typeof(LegalEntitiesController))
+                .AllowStatusCodes(HttpStatusCode.OK);
+
+            // Act
+            var account = await _tester.InvokeGetAsync<ResourceList>(callRequirements);
+
+            // Assert
+            Assert.IsNotNull(account.Data);
+ //           account.Data[0];
```
robinoconnor: this commented out part needs to be removed

src/SFA.DAS.EAS.Web.IntegrationTests/LegalEntitiesControllerTests/WhenIGetLegalEntitiesWithUnknownIds.cs
```diff
@@ -0,0 +1,40 @@
...
+        public async Task ThenTheStatusShouldBeNotFound_ByHashedId()
+        {
+            // Arrange
+            var callRequirements = new CallRequirements($"api/accounts/MADE*UP*ID/legalentities");
+
+            // Act
+            await _tester.InvokeGetAsync<ResourceList>(callRequirements);
+
+            // Assert
+            Assert.Pass("Verified we got http status NotFound");
```
DScuffle: We don't actually check for not found so I don't think we can assert a Pass here - previously we had a verification set up when creating the callRequirements object. In other tests we are doing something like account.ExpectStatusCodes(HttpStatusCode.OK);

iangledhill: They're extension methods on the return type from InvokeGetAsync<>() so if you assign the result to a local variable then you should see the available methods in intellisense.

src/SFA.DAS.EAS.Web.IntegrationTests/LegalEntitiesControllerTests/WhenIGetLegalEntitiesWithKnownIds.cs
```diff
@@ -0,0 +1,57 @@
...
+        [Test]
+        public async Task ThenTheStatusShouldBeFound_ByHashedAccountId()
+        {
+            var callRequirements = new CallRequirements($"api/accounts/{_employerAccount.HashedAccountId}/legalentities");
+
+            // Act
+            var account = await _tester.InvokeGetAsync<ResourceList>(callRequirements);
+
+            // Assert
+            Assert.IsNotNull(account.Data);
```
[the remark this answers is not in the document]
PeterDavidWilliams: Changes made - added count check and separated single legal entity and multiple legal entity tests out.

src/SFA.DAS.EAS.Web.IntegrationTests/LegalEntitiesControllerTests/WhenIGetLegalEntitiesWithUnknownIds.cs
PeterDavidWilliams: Changes made as requested.

PR #1504, opened 2018-12-07
Separate out the test setup from the test execution

iangledhill:
This isolates the IoC and db context used to create the test data for the setup from those that are used to execute the actual controller and method being tested.


scottywakefield:
I can't find where you are using a different IOC for data setup (likely my poor reading of the code). Also maybe we just need a subset of ioc configuration for the data setup. Not the full IOC used by the system under test, not worth holding up the PR for that though.

PR #1508, opened 2018-12-11
Populate cosmos account user collection

Pauljgraham:


src/SFA.DAS.EAS.Employer_Account.Database/Tables/JobHistory.sql
```diff
@@ -0,0 +1,5 @@
+﻿CREATE TABLE [dbo].[JobHistory]
+(
+	[Job] NVARCHAR(50) NOT NULL PRIMARY KEY,
```
lazcool: Having `Job` as the primary key means that the `JobHistory` table can only store a single run for each particular job.
Even though the `PopulateAccountUsersInCollectionJob` job is only meant to be run once, allowing the recording of multiple runs for a job would make the `JobHistory` table more useful for reuse.

Also, `NVARCHAR` isn't a great data type for a primary key. Would it be better to have a numeric identity column as the primary key?

src/SFA.DAS.EmployerAccounts.Jobs/Data/PopulateRepository.cs
```diff
@@ -0,0 +1,64 @@
...
+using System.Data;
+using System.Linq;
+using System.Threading.Tasks;
+using Dapper;
+using Microsoft.Extensions.Logging;
+using SFA.DAS.EmployerAccounts.Data;
+
+namespace SFA.DAS.EmployerAccounts.Jobs.Data
+{
+    public class PopulateRepository : IPopulateRepository
```
lazcool: What do you think of having [one repository per aggregate](https://docs.microsoft.com/en-us/dotnet/standard/microservices-architecture/microservice-ddd-cqrs-patterns/infrastructure-persistence-layer-design)?
So perhaps a [User|MembershipUser]Repository (or similar) and a JobHistoryRepository?
Also, I think it would be better if the repository was named after the entities involved, rather than a verb.

src/SFA.DAS.EmployerAccounts.Jobs/Data/PopulateRepository.cs
```diff
@@ -0,0 +1,64 @@
...
+    {
+        private readonly ILogger _logger;
+        private readonly EmployerAccountsDbContext _db;
+
+        public PopulateRepository(Lazy<EmployerAccountsDbContext> db, ILogger logger)
+        {
+            _logger = logger;
+            _db = db.Value;
```
lazcool: We could get more value out of the `Lazy<>` if we only referenced the `Value` just in time, rather than anticipatorily in the constructor.

src/SFA.DAS.EmployerAccounts.Jobs/Data/PopulateRepository.cs
```diff
@@ -0,0 +1,64 @@
...
+        }
+
+        public async Task<IEnumerable<MembershipUser>> GetAllAccountUsers()
+        {
+            _logger.LogInformation("Getting All Users to be migrated");
+
+            var result = await _db.Database.Connection.QueryAsync<MembershipUser>(
+                sql: @"SELECT M.UserId, U.UserRef, M.AccountId, M.RoleId FROM [employer_account].[Membership] M  
+                            LEFT JOIN [employer_account].[User] U ON M.UserId = U.Id 
+                            WHERE M.RoleId IN (1,2,3)",
```
lazcool: It would be nice if there were no magic numbers here (the RoleId's 1, 2 & 3), or at least a comment as to what those RoleId's represent.

src/SFA.DAS.EmployerAccounts.Jobs/Data/PopulateRepository.cs
```diff
@@ -0,0 +1,64 @@
...
+        {
+            _logger.LogInformation($"Checking if '{job}' has already run");
+
+            var param = new DynamicParameters();
+            param.Add("@job", job, DbType.String);
+
+            var result = await _db.Database.Connection.QueryAsync<int>(
+                sql: @"SELECT 1 FROM [dbo].[JobHistory] 
+                            WHERE Job = @job",
+                param: param,
```
lazcool: [Passing the parameters in an anonymous type](https://dapper-tutorial.net/parameter-anonymous) might be neater?

src/SFA.DAS.EmployerAccounts.Jobs/StartupJobs/CreateReadStoreDatabaseJob.cs
```diff
@@ -46,6 +49,8 @@ public async Task Run()
                 }
             };
 
+            _logger.LogInformation("Creating ReadStore database and collection if they don't exist");
+
             await _documentClient.CreateDatabaseIfNotExistsAsync(database);
```
lazcool: I think it would be really useful to log whether the database was created or not, which you can [tell from the response returned](https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.documents.client.documentclient.createdatabaseifnotexistsasync?view=azure-dotnet).

src/SFA.DAS.EmployerAccounts.Jobs/StartupJobs/CreateReadStoreDatabaseJob.cs
```diff
@@ -46,6 +49,8 @@ public async Task Run()
...
             await _documentClient.CreateDocumentCollectionIfNotExistsAsync(UriFactory.CreateDatabaseUri(database.Id), documentCollection);
```
lazcool: Similarly, it would be nice to log whether the collection was created or not.

src/SFA.DAS.EmployerAccounts.Jobs/StartupJobs/PopulateAccountUsersInCollectionJob.cs
```diff
@@ -0,0 +1,61 @@
...
+
+            var users = await _populateRepository.GetAllAccountUsers();
+
+            _logger.LogInformation("Migrating users into the read store"); 
+
+            var populateMessageId = Guid.NewGuid().ToString();
+
+            foreach (var user in users)
+            {
+                if (await _accountUsersRepository.CreateQuery().AnyAsync(x => x.AccountId == user.AccountId && x.UserRef == user.UserRef) == false)
```
lazcool:
```suggestion
                if (!await _accountUsersRepository.CreateQuery().AnyAsync(x => x.AccountId == user.AccountId && x.UserRef == user.UserRef))
```

src/SFA.DAS.EmployerAccounts.Jobs/StartupJobs/PopulateAccountUsersInCollectionJob.cs
```diff
@@ -0,0 +1,61 @@
...
+        {
+            _accountUsersRepository = accountUsersRepository;
+            _populateRepository = populateRepository;
+            _logger = logger;
+            _jobName = typeof(PopulateAccountUsersInCollectionJob).Name;
+        }
+
+        [NoAutomaticTrigger]
+        public async Task Run()
+        {
```
lazcool: Might be nice to log that the job has started.

src/SFA.DAS.EmployerAccounts/DependencyResolution/ConfigurationRegistry.cs
```diff
@@ -13,9 +13,9 @@ public ConfigurationRegistry()
             For<EmployerFinanceConfiguration>().Use(() => ConfigurationHelper.GetConfiguration<EmployerFinanceConfiguration>("SFA.DAS.EmployerFinance")).Singleton();
             //For<EmployerAccountsReadStoreConfiguration>().Use(() => ConfigurationHelper.GetConfiguration<EmployerAccountsReadStoreConfiguration>("SFA.DAS.EmployerAccounts.ReadStore")).Singleton();
```
lazcool: Either this line should be uncommented and the hard-coded values below removed from here and added into das-employer-config, or it could use the auto-configurator.

src/SFA.DAS.EAS.Employer_Account.Database/Tables/JobHistory.sql
Pauljgraham: Totally agree, will add an auto key and make the job name a secondary index.

src/SFA.DAS.EmployerAccounts.Jobs/Data/PopulateRepository.cs
Pauljgraham: Yes i see a lot of value in that. But the question then is where does this repository live. It's sole purpose is to populate the ReadStore

Pauljgraham: Have moved the IJobHistoryRepo into the application layer

src/SFA.DAS.EmployerAccounts.Jobs/StartupJobs/CreateReadStoreDatabaseJob.cs
Pauljgraham: Good idea, but the Database object doesn't specify a property which says whether this was created or not. Well not that I can see

src/SFA.DAS.EmployerAccounts.Jobs/StartupJobs/CreateReadStoreDatabaseJob.cs
Pauljgraham: As above, we can't see if this created the collection or just found it

src/SFA.DAS.EmployerAccounts/DependencyResolution/ConfigurationRegistry.cs
Pauljgraham: AutoConfig will do this

src/SFA.DAS.EmployerAccounts.Jobs/StartupJobs/CreateReadStoreDatabaseJob.cs
lazcool: Added logging to say if the database and document collection were actually created or not.

src/SFA.DAS.EmployerAccounts/DependencyResolution/ConfigurationRegistry.cs
lazcool: If we use auto config, then don't we need to do something along the lines of...

```
IncludeRegistry<AutoConfigurationRegistry>();
For<EmployerAccountsReadStoreConfiguration>().Use(c => c.GetInstance<ITableStorageConfigurationService>().Get<EmployerAccountsReadStoreConfiguration>());
```
Also, `ConfigurationRegistry` is used in 5 different places. Do we need access to EmployerAccountsReadStoreConfiguration in all those places? If not, we could only make the config available when it's actually required.

elliotritchie:
Merged into PR #1514.

PR #1511, opened 2018-12-17
Separated api call asserts from set up

iangledhill:
I've taken the expectations out of the call requirements and moved them to checking the call response. This makes the tests look more conventional.


iangledhill:
@scottywakefield 
Regarding making fluent - yeah, they were fluent before but assertions are not usually fluent so I stuck to the normal convention.

I made the extension methods to keep the class uncluttered as you say. I wanted the call response to be just a statement of what had happened, with no behaviour. We will want to extend the number and type of asserts and I figured this would be easier as ext methods.

PR #1512, opened 2018-12-17
Changed code to use different list for integration test

robinoconnor:


src/DefaultTargets.fsx
```diff
@@ -463,6 +464,29 @@ Target "Run Acceptance Tests" (fun _ ->
...
+    trace "Run Integration Tests"
+
+    let mutable shouldRunTests = false
+
+    let testDlls = !! ("./**/bin/" + testDirectory + "/*.IntegrationTests.dll")
+
+    for testDll in testDlls do
+        shouldRunTests <- true
+
+    if shouldRunTests then
```
iangledhill: I don't know the syntax of the fake script but this _looks_ like shouldRunTests might always be true, and so could be removed.

robinoconnor: I've just copied this for now as it's doing the exact same thing as the acceptance tests. Given i'm not a pro at F# either I'd rather keep it in for now and then create a ticket to look at the whole script later on as its not stopping issue for now.

robinoconnor: Just taken a look at it seems to be used as a trigger to run the tests if any are present and not to try to run the tests if none are found.

PR #1515, opened 2018-12-19
Added portal redirect url to accounts HMRC request

robinoconnor:


src/SFA.DAS.EAS.Web/Controllers/EmployerAccountController.cs
```diff
@@ -25,7 +25,8 @@ public ActionResult Gateway()
         [Route("gatewayResponse")]
         public ActionResult GateWayResponse()
         {
-            return Redirect(Url.EmployerAccountsAction("gatewayResponse"));
+            return Redirect(Url.EmployerAccountsAction($"gatewayResponse?{Request.QueryString}"));
```
scottywakefield: I'm guessing that the Request.QueryString is already URL encoded and so doesn't need any checks of that nature.

robinoconnor: that is a good point. I should have used Request.url.query instead but the QueryString.tostring worked on this occasion (probably because there is only one param)

robinoconnor: This has now been changed to the correct value which is encoded to URI specification

src/SFA.DAS.EAS.Web/Controllers/EmployerAccountController.cs
```diff
@@ -25,7 +25,7 @@ public ActionResult Gateway()
         [Route("gatewayResponse")]
         public ActionResult GateWayResponse()
         {
-            return Redirect(Url.EmployerAccountsAction($"gatewayResponse?{Request.QueryString}"));
+            return Redirect(Url.EmployerAccountsAction($"gatewayResponse{Request?.Url?.Query}"));
```
iangledhill: Can Url and Query actually ever be null?

robinoconnor: its just the Query from the looks of it. i'll update the code now

robinoconnor: Url can be null as its a public variable that can be set and given its a complex type I'm guessing they have to cover for it being null. Not sure if it could ever be null be default however with regards to this action.

PR #1520, opened 2019-01-03
Aml 2986 duplicate organisations

iangledhill:
The purpose of this change is to make the "remove organisation" functionality work _directly_ against organisations (aka legal entity) rather than via agreement id. 
This fixes several problems:
1. duplicate organisations occurring when the account has multiple active agreements
2. when an agreement was previously removed *all* agreements linked to the same organisation were removed but *only* the agreement selected from the screen went through the pipeline (i.e. had audit and events created).


DannyGB:

PR #1524, opened 2019-01-09
Aml 3097 fix nsb event errors

iangledhill:
This change does two things:
Firstly, it uses convention for identifying nservicebus commands and events (rather than using IEvent and ICommand). This is the recommended way of doing things and is the way adopted by SFA.

The second is the more important fix - it removes the dependency on the base abstract class Event, which is causing errors in prod. 

In order to achieve these it was necessary to update the SFA UoW and SFA NSB nuget packages. This was because the older nuget packages we were using had generic class constraints that specified the classes and interfaces we are removing. These nuget packages also had some additional breaking changes related to publishing events (specifically, the signature on UnitOfWorkContext.AddEvent has changed), which have been addressed.


DannyGB:

PR #1525, opened 2019-01-10
AML-3077 Clear local storage when we navigate to the setup account view

DScuffle:
Welcome wizard works by using javascript localStorage - this is never cleared down. I have set this to clear any time the users tries to setup a new account.


DScuffle:
Scrap this - wrong work item number used - recreated the PR

PR #1528, opened 2019-01-17
aml-3066 Fix Agreement Signed Step in Welcome Wizard

DScuffle:
Slight modification to the Welcome Wizard. 

Step 4 now gives you a count of signed agreements in it's done state and  there is an additional footer of info that lets you know if you have any that are unsigned.

In order to do that I have had to add the Signed Agreement Count onto the view model.


DannyGB:
The original author has left and the PR is over 12 months old and will never be merged into master

PR #1531, opened 2019-01-31
Con 82 two different apprentices with the same name being aggregated

Ashley-Bass:
Apprentices with the same name were being aggregated. Changed sorting and grouping to be by ULN instead of name. Also brought ULN up through from SQL to Orchestrator for grouping there.

### DIFF
--- a/src/SFA.DAS.EmployerFinance.Web/Orchestrators/EmployerAccountTransactionsOrchestrator.cs
+++ b/src/SFA.DAS.EmployerFinance.Web/Orchestrators/EmployerAccountTransactionsOrchestrator.cs
@@ -199,7 +199,7 @@ namespace SFA.DAS.EmployerFinance.Web.Orchestrators
                     ExternalUserId = externalUserId
                 });
 
-                var apprenticePaymentGroups = data.Transactions.GroupBy(x => new { x.ApprenticeULN,x.ApprenticeName });
+                var apprenticePaymentGroups = data.Transactions.GroupBy(x => new { x.ApprenticeULN });
 
                 var paymentSummaries = apprenticePaymentGroups.Select(pg =>
                 {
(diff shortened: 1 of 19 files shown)


src/SFA.DAS.EmployerFinance.Web/Views/EmployerAccountTransactions/CoursePaymentSummary.cshtml
```diff
@@ -54,40 +54,44 @@
...
-                        <td class="numeric" data-label="Paid from levy">@coursePaymentSummary.LevyPaymentAmount.ToString("C2", culture)</td>
-                        <td class="numeric highlight" data-label="Government contribution">@coursePaymentSummary.SFACoInvestmentAmount.ToString("C2", culture)</td>
-                        <td class="numeric highlight" data-label="Your contribution">@coursePaymentSummary.EmployerCoInvestmentAmount.ToString("C2", culture)</td>
-                        <td class="numeric" data-label="Amount">@coursePaymentSummary.TotalAmount.ToString("C2", culture)</td>
-                    </tr>
+                <tr class="responsive-tr">
+                    @if (string.IsNullOrEmpty(@coursePaymentSummary.ApprenticeName))
+                    {
+                        <td data-label="Apprentice"><em>Not available</em></td>
+                        <td data-label="Unique learner number (ULN)"><em>Not available</em></td>
```
cofaulco: If the ULN is available but Apprentice name is not, we are showing both as "Not available"

src/SFA.DAS.EmployerFinance.Web/Orchestrators/EmployerAccountTransactionsOrchestrator.cs
```diff
@@ -199,7 +199,7 @@ public EmployerAccountTransactionsOrchestrator(IMediator mediator, ICurrentDateT
                     ExternalUserId = externalUserId
                 });
 
-                var apprenticePaymentGroups = data.Transactions.GroupBy(x => new { x.ApprenticeName, x.ApprenticeNINumber });
+                var apprenticePaymentGroups = data.Transactions.GroupBy(x => new { x.ApprenticeULN,x.ApprenticeName });
```
cofaulco: If ApprenticeUln is unique, then is there any value to the secondary grouping of apprentice name

Ashley-Bass: No, but I knew of no other way of keeping the name to be passed into line 210

PR #1537, opened 2019-02-08
AML-3291:update contents as per prototype in ticket. Note: not tested…

Winsto:
Note: not confirmed in local development environment.


soharadesign:
Testing currently impossible as the builds aren't working locally. _This will be rectified as part of a separate ticket_

PR #1540, opened 2019-02-08
CON-88 Organisation should be hidden when it has ongoing apprentices

Ashley-Bass:
Created a fix that hides all organisations that aren't removable and just shows a message stating no organisations can be removed


src/SFA.DAS.EmployerAccounts.Web/Views/EmployerAgreement/GetOrganisationsToRemove.cshtml
```diff
@@ -25,32 +25,28 @@
                     <th scope="col"><span class="vh">Status</span></th>
             </thead>
             <tbody>
+                @{
```
cofaulco: Can we change this code to check the Agreements Collection for any orgs that can be removed rather than this method of setting a count?

cofaulco:
Following discussions, closing this as the issue should be fixed by a data fix and the code change isn't fit for the purpose of the story

PR #1542, opened 2019-02-11
CON-199 Removed RoleId as was marked obsolete

Ashley-Bass:
Completely removed RoleId from the code in favour of using just Role. Some tests were updated to match the new variables. Ran locally and all tests passed fine and no functionality seems to be compromised


src/SFA.DAS.EAS.Account.Api.Types/TeamMemberViewModel.cs
```diff
@@ -11,7 +11,7 @@ public class TeamMemberViewModel
 
         public string Email { get; set; }
 
-        public string Role { get; set; }
+        public string Role{ get; set; }
```
cofaulco: Spacing

src/SFA.DAS.EAS.Employer_Account.Database/StoredProcedures/AcceptInvitation.sql
```diff
@@ -2,7 +2,7 @@
 (
 	@email NVARCHAR(255),
 	@accountId BIGINT,
-	@roleId TINYINT
+	@Role TINYINT
```
cofaulco: casing consistency

PR #1543, opened 2019-02-12
Con 101 membership model has incorrect role logic

Ashley-Bass:
Fixed role logic so that no nulls can occur


Ashley-Bass:
Closing branch as has been overwrittten by CON-199

PR #1547, opened 2019-02-14
Con 98 get team member throwing exception when member does not exist

Ashley-Bass:


src/SFA.DAS.EmployerAccounts.Web/Orchestrators/EmployerTeamOrchestrator.cs
```diff
@@ -274,11 +274,20 @@ private async Task<OrchestratorResponse<TeamMember>> GetTeamMember(string hashed
                 Email = email,
                 OnlyIfMemberIsActive = onlyIfMemberIsActive
             });
-
-            return new OrchestratorResponse<TeamMember>
+            if (response.TeamMember.Role == Role.None)
```
cofaulco: Is it possible for a valid team member to exist with a role of None?

PR #1551, opened 2019-02-21
Aml 3315 update readme

Winsto:


Winsto:
@cofaulco I might update this a bit more. It seems there can be extra shenanigans around compute emulator running the cloud service.

DannyGB:
The original author has left and the PR is 12 months old and will never be merged into master

PR #1554, opened 2019-02-26
Aml 3245 expired funds bingo box

robinoconnor:
Wired in the expired funds bingo box into existing code base. We can upgrade the packages to use the latest Http client code but this requires a update of the notification api client as well so will be done in a different branch


robinoconnor:
To get this code running locally you will need the config which has been added under this PR: https://github.com/SkillsFundingAgency/das-employer-config/pull/216

src/SFA.DAS.EmployerFinance.Web.UnitTests/Controllers/EmployerAccountTransactionsControllerTests/WhenIViewFinanceDashboard.cs
```diff
@@ -46,24 +59,26 @@ public void Arrange()
         }
 
         [Test]
-        public async Task ThenTransactionsAreRetrievedForTheAccount()
+        public async Task ThenThAccountHashedIdIsReturned()
```
cofaulco: Typo in Unit test name

src/SFA.DAS.EmployerFinance/Models/ExpiringFunds/ExpiringAccountFunds.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace SFA.DAS.EmployerFinance.Models.ExpiringFunds
+{
+    public class ExpiringAccountFunds
+    {
+        public long? AccountId { get; set; }
```
cofaulco: Should accountId be nullable as we are searching for these funds based on it

src/SFA.DAS.EmployerFinance/Queries/GetAccountFinanceOverview/GetAccountFinanceOverviewQueryHandler.cs
```diff
@@ -0,0 +1,81 @@
...
+                    AccountId = query.AccountId
+                };
+            }
+
+            return new GetAccountFinanceOverviewResponse
+            {
+                AccountId = query.AccountId,
+                CurrentFunds = currentBalance,
+                ExpiringFundsExpiryDate = earliestFundsToExpire.PayrollDate,
+                ExpiringFundsAmount = earliestFundsToExpire.Amount
```
cofaulco: Expiring funds is nullable but is only assigned to by setting it as a none nullable long. Should we make these consistent?

src/SFA.DAS.EmployerFinance/Queries/GetAccountFinanceOverview/GetAccountFinanceOverviewQueryHandler.cs
```diff
@@ -0,0 +1,81 @@
...
+            ILog logger)
+        {
+            _dasForecastingService = dasForecastingService;
+            _levyService = levyService;
+            _logger = logger;
+        }
+
+        public async Task<GetAccountFinanceOverviewResponse> Handle(GetAccountFinanceOverviewQuery query)
+        {
+            if (!query.AccountId.HasValue)
```
cofaulco: Should this not be done via a validator?

src/SFA.DAS.EmployerFinance/Services/DasForecastingService.cs
```diff
@@ -0,0 +1,80 @@
...
+                    case 408 : _logger.Error(ex, $"Request sent to forecast API for account ID: {accountId} timed out");
+                        break;
+                    case 429 : _logger.Error(ex, $"To many requests sent to forecast API for account ID: {accountId}");
+                        break;
+                    case 500 : _logger.Error(ex, $"Forecast API reported internal Server error for account ID: {accountId}");
+                        break;
+                    case 503 : _logger.Error(ex, "Forecast API is unavailable");
+                        break;
+
+                    default: throw;
```
cofaulco: Nothing handles this throw. So error page will be displayed

robinoconnor: That's correct as we handle all of the expected errors that could occur but do not handle an unexpected errors. I'll ask pritesh what we should do in that scenario.

src/SFA.DAS.EmployerFinance/Queries/GetAccountFinanceOverview/GetAccountFinanceOverviewQueryHandler.cs
robinoconnor: good point, no idea why i didn't use one. I'll change this now.

robinoconnor: Actually I do after looking at it. The validation is handled in the MVC mapping via a filter. However I'm not sure if this could be called via an internal process so need to get some feedback around that first before I change anything.

robinoconnor: I think that's why I didn't do fully validation on it

PR #1555, opened 2019-02-26
Handle StructureMap error on disposal when httpContext is null.

Stephek:


bucky-uk (commented):
@cofaulco @MichaelYoung1981 I think this can be closed now 😉

PR #1569, opened 2019-03-15
AML-3430 Added funds in stored proc

tom-gough:


src/SFA.DAS.EAS.Employer_Financial.Database/StoredProcedures/GetFundsIn.sql
```diff
@@ -0,0 +1,16 @@
+﻿CREATE PROCEDURE [employer_financial].[GetFundsIn]
```
elliotritchie: It might be worth renaming the repo and sproc to match the type it returns i.e. `ILevyFundsInRepository` & `GetLevyFundsIn`.

PR #1572, opened 2019-03-19
Aml-3433 Write Expired Funds

lazcool:
`ExpiredFundsRepository.Create()` has been tested manually.

Contains the '[add funds in PR](https://github.com/SkillsFundingAgency/das-employerapprenticeshipsservice/pull/1569)'.

### DIFF
--- a/src/SFA.DAS.EAS.Employer_Financial.Database/StoredProcedures/CreateExpiredFunds.sql
+++ b/src/SFA.DAS.EAS.Employer_Financial.Database/StoredProcedures/CreateExpiredFunds.sql
@@ -1,5 +1,5 @@
 ﻿CREATE PROCEDURE [employer_financial].[CreateExpiredFunds]
-	@accountId BIGINT NOT NULL,
+	@accountId BIGINT,
 	@expiredFunds [employer_financial].[ExpiredFundsTable] READONLY
 AS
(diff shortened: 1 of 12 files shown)


src/SFA.DAS.EAS.Employer_Financial.Database/StoredProcedures/CreateExpiredFunds.sql
```diff
@@ -0,0 +1,7 @@
+﻿CREATE PROCEDURE [employer_financial].[CreateExpiredFunds]
+	@accountId BIGINT,
+	@expiredFunds [employer_financial].[ExpiredFundsTable] READONLY
+AS
+	INSERT [employer_financial].[TransactionLine] (AccountId, DateCreated, TransactionDate, TransactionType, Amount)
```
elliotritchie: It might be worth adding a unique index to the `TransactionLine` table on the `TransactionDate` and `TransactionType` columns where `TransactionType = 5` to ensure we never get any dupes.

lazcool: Added, with columns ordered to help speed up the `GetExpiredFunds` stored proc.

PR #1573, opened 2019-03-19
AML-3444 AML-3431 funds out spike

tom-gough:


src/SFA.DAS.EAS.Employer_Financial.Database/StoredProcedures/GetFundsOut.sql
```diff
@@ -0,0 +1,14 @@
+﻿CREATE PROCEDURE [employer_financial].[GetFundsOut]
```
elliotritchie: It might be worth renaming the repo and sproc to match the type it returns i.e. `IPaymentFundsOutRepository` & `GetPaymentFundsOut`.

src/SFA.DAS.EAS.Employer_Financial.Database/StoredProcedures/GetFundsOut.sql
```diff
@@ -0,0 +1,14 @@
...
+	@AccountId bigint
+AS
+SELECT	periodEnd.CalendarPeriodMonth AS CalendarPeriodMonth,
+			periodEnd.CalendarPeriodYear AS CalendarPeriodYear,
+			SUM(transactionLine.[Amount]) AS FundsOut
+	FROM [employer_financial].[TransactionLine] transactionLine
+	INNER JOIN [employer_financial].[PeriodEnd] periodEnd
+	ON periodEnd.PeriodEndId = transactionLine.PeriodEnd
+	
+	WHERE transactionLine.AccountId = @AccountId
```
elliotritchie: I think this query will need updating to only include transfer transaction line amounts for a sender and not a receiver.

tom-gough: Had a conversation with Elliot. We don't need to update this as the transaction lines for transfers handle the changes in levy amount (via levy being transferred and corresponding payments) correctly. To filter these out is un-necessary complexity

PR #1575, opened 2019-03-20
Aml-3446 Add Created To DeletedPayeSchemeEvent

lazcool:


src/SFA.DAS.EmployerAccounts.UnitTests/Commands/RemovePayeFromAccountTests/WhenIRemoveAPayeSchemeFromAnAccount.cs
```diff
@@ -181,6 +181,7 @@ public async Task ThenAMessageIsQueuedForPayeSchemeRemoved()
             message.PayeRef.Should().Be(PayeScheme);
             message.UserName.Should().Be(UserName);
             message.UserRef.Should().Be(UserRef);
+            Assert.IsTrue(DateTime.UtcNow - message.Created < TimeSpan.FromMinutes(1));
```
lazcool: I tried several combinations of FluentAssertion's `DateTime` assertions, such as `BeWithin()`, but none of them seemed to work as I'd expect, so either they have issues, or (more likely) I'm not understanding them, in which case I thought it best to avoid them as they're confusing.

PR #1582, opened 2019-03-22
AML-3456 Orchestrator error response bug fix

elliotritchie:


src/SFA.DAS.EmployerFinance.Web/Controllers/BaseController.cs
```diff
@@ -35,38 +36,23 @@ protected override ViewResult View(string viewName, string masterName, object mo
...
             }
 
-            if (orchestratorResponse.Status == HttpStatusCode.Unauthorized)
+            if (orchestratorResponse.Status >= HttpStatusCode.BadRequest)
             {
-                var accountId = Request.Params[ControllerConstants.AccountHashedIdRouteKeyName];
-
-                if (accountId != null)
-                {
-                    ViewBag.AccountId = accountId;
```
lazcool: Looks like the old access denied view might have needed the AccountId. The new one doesn't, so we don't need to supply it, but I'm wondering if we've lost something that we should be in the new view?

lazcool: Per @elliotritchie, the old view used the layout, so required the account id, whereas the new one doesn't (it's purely static), so doesn't need the AccountId.

PR #1583, opened 2019-03-22
[CON-329] Add separate command handler to process period end to avoid…

cofaulco:
… single huge outbox transaction for multiple periods

### DIFF
--- a/src/SFA.DAS.EmployerFinance.MessageHandlers/CommandHandlers/ProcessPeriodEndPaymentsCommandHandler.cs
+++ b/src/SFA.DAS.EmployerFinance.MessageHandlers/CommandHandlers/ProcessPeriodEndPaymentsCommandHandler.cs
@@ -29,11 +29,13 @@ namespace SFA.DAS.EmployerFinance.MessageHandlers.CommandHandlers
             {
                 _logger.Info($"Creating payment queue message for account ID: '{account.Id}' period end ref: '{message.PeriodEndRef}'");
 
-                tasks.Add(context.SendLocal<ImportAccountPaymentsCommand>(c =>
-                {
-                    c.AccountId = account.Id;
-                    c.PeriodEndRef = message.PeriodEndRef;
-                }));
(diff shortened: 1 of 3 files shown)


src/SFA.DAS.EmployerFinance.MessageHandlers/CommandHandlers/ProcessPeriodEndPaymentsCommandHandler.cs
```diff
@@ -0,0 +1,42 @@
...
+        {
+            var response = await _mediator.SendAsync(new GetAllEmployerAccountsRequest());
+
+            var tasks = new List<Task>();
+
+            foreach (var account in response.Accounts)
+            {
+                _logger.Info($"Creating payment queue message for account ID: '{account.Id}' period end ref: '{message.PeriodEndRef}'");
+
+                tasks.Add(context.SendLocal<ImportAccountPaymentsCommand>(c =>
```
elliotritchie: There'll potentially be tens of thousands of messages (one for each account) so the outbox might be a bottleneck here so it might be worth doing something along the lines of:

```c#
var sendOptions = new SendOptions();

sendOptions.RouteToThisEndpoint();
sendOptions.RequireImmediateDispatch(); // Circumvent sender outbox
sendOptions.SetMessageId($"{message.PeriodEndRef}-{account.Id}"); // Allow receiver outbox to de-dupe

context.Send(new ImportAccountPaymentsCommand { PeriodEndRef = message.PeriodEndRef, AccountId = account.Id }, sendOptions);
```

elliotritchie: @cofaulco Updated my comment, probably worth including the name of the message in the ID to ensure it's unique.

cofaulco:
Closing as merged into CON-63 in order to enable testing of CON-63 as this fix was preventing that

PR #1595, opened 2019-04-05
Added Confirm Who You Are page

Howesy77:
Work for Tasks:-

AML-3476, AML-3289

Fix for Bug:-

AML-3511

### DIFF
--- a/src/SFA.DAS.EmployerAccounts.Web/Controllers/HomeController.cs
+++ b/src/SFA.DAS.EmployerAccounts.Web/Controllers/HomeController.cs
@@ -127,6 +127,38 @@ namespace SFA.DAS.EmployerAccounts.Web.Controllers
         }
 
         [HttpGet]
+        [Route("confirmWhoYouAre")]
+        public ActionResult ConfirmWhoYouAre()
+        {
+            var model = new
+            {
+                HideHeaderSignInLink = true
+            };
+
+            return View(model);
+        }
+
+        [HttpPost]
+        [Route("confirmWhoYouAre")]
+        public ActionResult ConfirmWhoYouAre(int? choice)
+        {
+            switch (choice ?? 0)
+            {
+                case 1: return RedirectToAction(ControllerConstants.ConfirmWhoYouAre); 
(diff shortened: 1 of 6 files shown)


src/SFA.DAS.EmployerAccounts.Web/Controllers/HomeController.cs
```diff
@@ -140,9 +172,21 @@ public ActionResult WhatYoullNeed()
 
         [HttpPost]
         [Route("whatYoullNeed")]
-        public ActionResult WhatYoullNeedPost()
+        public ActionResult WhatYoullNeed(int? choice)
         {
-            return RedirectToAction(ControllerConstants.RegisterUserActionName);
+            switch (choice ?? 0)
+            {
+                case 2: return RedirectToAction(ControllerConstants.RegisterUserActionName);
```
Winsto: Does the case where the user has answered 'No' not also need to return as error?

Howesy77: It will as anything other than yes uses the default route.

Howesy77: Not sure about the redirect as it’s not yet exposed in the current journey so I assume this can be looked into once we use the page.

Winsto: > It will as anything other than yes uses the default route.

Ahh, sorry I assumed '2' was 'No' thinking '1', '2' going down in list, but 'obviously' that list goes in reverse order.

Winsto: > Not sure about the redirect as it’s not yet exposed in the current journey so I assume this can be looked into once we use the page.

Yes, I did wonder about that, but can it be tested (via direct URL) without the re-direct?

Winsto: Thinking about it, we should not have to put new page re-directs in; iirc.

PR #1598, opened 2019-04-11
AML-3461 Added expired funds acceptance tests

elliotritchie:


src/SFA.DAS.EAS.Employer_Financial.Database/StoredProcedures/DeleteSubmissions_ByEmpRef.sql
```diff
@@ -13,14 +13,14 @@ AS
 
 	DELETE	
 	FROM	employer_financial.TransactionLine 
-	WHERE	SubmissionId IN (SELECT SubmissionId FROM @SubmissionIdsInUseByEmpRef);
+	WHERE EmpRef = @empRef
 
 	DELETE 
 	FROM	employer_financial.LevyDeclarationTopup
 	WHERE	SubmissionId IN (SELECT SubmissionId FROM @SubmissionIdsInUseByEmpRef);
```
lazcool: We could select directly from the LevyDeclaration table now, and remove the table variable, now that we only need the results in one place.

src/SFA.DAS.EmployerFinance.AcceptanceTests/Steps/FinanceSteps.cs
```diff
@@ -22,6 +34,76 @@ public FinanceSteps(IObjectContainer objectContainer, ObjectContext objectContex
...
+            await _objectContainer.ScopeAsync(async c =>
+            {
+                var transactionRepository = c.Resolve<ITransactionRepository>();
+                var timeout = Debugger.IsAttached ? 10 * 60 * 1000 : StepTimeout;
+                var cancellationTokenSource = new CancellationTokenSource(timeout);
+                var isComplete = await transactionRepository.WaitForAllTransactionLinesInDatabase(account, cancellationTokenSource.Token);
+
+                if (!isComplete)
+                {
+                    throw new Exception($"The transactions have not been completely loaded within the allowed time ({timeout} msecs). Either they are still loading or something has failed.");
```
lazcool: We could instead explicitly fail the test using `Assert.Fail()`.

src/SFA.DAS.EmployerFinance.AcceptanceTests/TestRepositories/TestTransactionRepository.cs
```diff
@@ -28,10 +31,53 @@ public class TestTransactionRepository : BaseRepository, ITestTransactionReposit
...
-                sql: "SELECT COALESCE(MAX(AccountId), 0) FROM [employer_financial].LevyDeclaration",
+                sql: @"
+                    SELECT COALESCE(MAX(a.AccountId), 0)
+                    FROM
+                    (
+	                    SELECT AccountId FROM [employer_financial].LevyDeclaration
+                        UNION
+	                    SELECT AccountId FROM [employer_financial].Payment
+                        UNION
+	                    SELECT AccountId FROM [employer_financial].TransactionLine
```
lazcool: Couldn't we get the max AccountId just by looking at TransactionLine?

elliotritchie: I don't think so, it looks like tests can fail halfway through (e.g. `LevyDeclaration` data has been populated for the test account but not corresponding `TransactionLine` data). Ideally it'd come from an `Accounts` table though!

src/SFA.DAS.EmployerFinance.AcceptanceTests/Steps/FinanceSteps.cs
elliotritchie: Done

src/SFA.DAS.EAS.Employer_Financial.Database/StoredProcedures/DeleteSubmissions_ByEmpRef.sql
elliotritchie: I've reverted the sproc to how it used to work to ensure it doesn't break the other tests and just added `OR EmpRef = @empRef`.

PR #1609, opened 2019-04-17
Aml 3484 checkdetails new link

Winsto:
Creating PR to feature branch for now.


Winsto:
Do not review yet! We are changing from using config to using feature toggles.

Winsto:
Switched over to FeatureToggle, so review away :0)

PR #1616, opened 2019-04-25
CON-451 Create WebJob Host

lazcool:


src/SFA.DAS.EAS.Portal.Client/Application/Queries/GetAccountQuery.cs
```diff
@@ -0,0 +1,25 @@
...
+    internal class GetAccountQuery
+    {
+        private readonly IAccountsReadOnlyRepository _accountsReadOnlyRepository;
+
+        public GetAccountQuery(IAccountsReadOnlyRepository accountsReadOnlyRepository)
+        {
+            _accountsReadOnlyRepository = accountsReadOnlyRepository;
+        }
+
+        public async Task<AccountDto> Get(long accountId, CancellationToken cancellationToken = default)
```
Stephek: Be better to stream the Account ReadModel direct from the cosmos database.  Add as tech debt in 454.

src/SFA.DAS.EAS.Portal.Client/Data/AccountsReadOnlyRepository.cs
```diff
@@ -0,0 +1,14 @@
+using Microsoft.Azure.Documents;
+using SFA.DAS.CosmosDb;
+using SFA.DAS.EAS.Portal.Client.Models.Concrete;
+
+namespace SFA.DAS.EAS.Portal.Client.Data
+{
+    internal class AccountsReadOnlyRepository : ReadOnlyDocumentRepository<AccountDto>, IAccountsReadOnlyRepository
```
Stephek: As per previous comment about streaming the ReadModel direct.

src/SFA.DAS.EAS.Portal.Client/Data/DocumentClientFactory.cs
```diff
@@ -0,0 +1,31 @@
...
+            _cosmosDbConfiguration = cosmosDbConfiguration;
+        }
+
+        public IDocumentClient CreateDocumentClient()
+        {
+            var connectionPolicy = new ConnectionPolicy
+            {
+                RetryOptions =
+                {
+                    MaxRetryAttemptsOnThrottledRequests = 2,
```
Stephek: Need to put these in config.

src/SFA.DAS.EAS.Portal.Client/IPortalClient.cs
```diff
@@ -0,0 +1,11 @@
+using System.Threading;
+using System.Threading.Tasks;
+using SFA.DAS.EAS.Portal.Client.Models.Concrete;
+
+namespace SFA.DAS.EAS.Portal.Client
+{
+    public interface IPortalClient
+    {
+        Task<AccountDto> GetAccount(long accountId, CancellationToken cancellationToken = default);
```
Stephek: Return the Account read model.   Add as tech debt in 454.

src/SFA.DAS.EAS.Portal.Client/Models/Concrete/AccountDto.cs
```diff
@@ -0,0 +1,28 @@
+using System;
+using System.Collections.Generic;
+using Newtonsoft.Json;
+
+namespace SFA.DAS.EAS.Portal.Client.Models.Concrete
+{
+    public class AccountDto : IAccountDto
```
Stephek: No need for an interface on the ReadModel.   Add as tech debt in 454.
Should drop the Hungarian suffix.  The namespace gives the context.

src/SFA.DAS.EAS.Portal.Client/Models/Concrete/AccountDto.cs
```diff
@@ -0,0 +1,28 @@
...
+        //todo: don't return accountId or Deleted
+        //todo: we now have a mix'n'match return of concrete containing ienumerable of interfaces!
+        
+        [JsonProperty("accountId")]
+        public virtual long AccountId { get; private set; }
+
+        //todo: return List in dto's, not enumerable (so consumer can add to them without copying - return most specific type)
+        // https://stackoverflow.com/questions/15490633/why-cant-i-use-a-compatible-concrete-type-when-implementing-an-interface
+        [JsonProperty("accountLegalEntities")]
+        public virtual IEnumerable<IAccountLegalEntityDto> AccountLegalEntities { get; private set; }
```
Stephek: These are Organisations.  That is what the screens / business requirements are.  We should be using the Account model that Ashley created which mirrors the business terminology.  Add as tech debt in 454.

src/SFA.DAS.EAS.Portal.Client/Models/Concrete/AccountLegalEntityDto.cs
```diff
@@ -0,0 +1,22 @@
...
+        }
+
+        [JsonProperty("accountLegalEntityId")]
+        public virtual long AccountLegalEntityId { get; private set; }
+
+        [JsonProperty("legalEntityName")]
+        public virtual string LegalEntityName { get; private set; }
+        
+        [JsonProperty("reservedFundings")]
+        public virtual IEnumerable<IReservedFundingDto> ReservedFundings { get; private set; }
```
Stephek: These are Reservations.  That is the domain the events that create them come from. Already in Ashley's Account read model.

src/SFA.DAS.EAS.Portal.Client/Models/Concrete/ReservedFundingDto.cs
```diff
@@ -0,0 +1,25 @@
+using System;
+using Newtonsoft.Json;
+
+namespace SFA.DAS.EAS.Portal.Client.Models.Concrete
+{
+    public class ReservedFundingDto : IReservedFundingDto
```
Stephek: As previous should be as per Ashley's ReadModel. Add as tech debt in 454.

src/SFA.DAS.EAS.Portal.Client/Models/IAccountDto.cs
```diff
@@ -0,0 +1,12 @@
+using System;
+using System.Collections.Generic;
+
+namespace SFA.DAS.EAS.Portal.Client.Models
+{
+    public interface IAccountDto
```
Stephek: No need for this interface on the read model.   There is no requirement for the read model to be polymorphic.

src/SFA.DAS.EAS.Portal/Application/Commands/AddReservationCommand.cs
```diff
@@ -0,0 +1,52 @@
...
+
+        public AddReservationCommand(IAccountsRepository accountsRepository, ILogger<AddReservationCommand> logger)
+        {
+            _accountsRepository = accountsRepository;
+            _logger = logger;
+        }
+
+        public async Task Execute(ReservationCreatedEvent reservedFunding, string messageId, CancellationToken cancellationToken = default)
+        {
+            _logger.LogInformation("Executing AddReservationCommand");
```
Stephek: No need to log this.  This type of logging could be added as a decorator of this method (if it was from an interface).  That would prevent mixing of concerns and make unit testing easier. Add as tech debt in 454.

src/SFA.DAS.EAS.Portal/Application/Commands/AddReservationCommand.cs
```diff
@@ -0,0 +1,52 @@
...
+
+            var account = await _accountsRepository
```
Stephek: Maybe hide this query behind a service interface. Add as tech debt in 454.

src/SFA.DAS.EAS.Portal/Application/Commands/AddReservationCommand.cs
```diff
@@ -0,0 +1,52 @@
...
+                .CreateQuery()
+                .SingleOrDefaultAsync(a => a.AccountId == reservedFunding.AccountId, cancellationToken);
+
+            if (account == null)
+            {
+                //first time we've had *any* event relating to this account
+                account = new Account(reservedFunding.AccountId, reservedFunding.AccountLegalEntityId, reservedFunding.AccountLegalEntityName,
```
Stephek: Mixing account creation and reservation handling together.  Better to keep separate. Add as tech debt in 454.

src/SFA.DAS.EAS.Portal/Database/DocumentClientFactory.cs
```diff
@@ -0,0 +1,34 @@
...
+            _options = options;
+        }
+
+        public IDocumentClient CreateDocumentClient()
+        {
+            var connectionPolicy = new ConnectionPolicy
+            {
+                RetryOptions =
+                {
+                    MaxRetryAttemptsOnThrottledRequests = 4,
```
Stephek: Should be from config.

src/SFA.DAS.EAS.Portal/Database/Models/Account.cs
```diff
@@ -0,0 +1,116 @@
...
+    public class Account : Document, IAccountDto
+    {
+        [JsonProperty("accountId")]
+        public long AccountId { get; private set; }
+
+        [JsonProperty("accountLegalEntities")]
+        public IEnumerable<IAccountLegalEntityDto> AccountLegalEntities => _accountLegalEntities;
+
+        [JsonProperty("outboxData")]
+        public IEnumerable<OutboxMessage> OutboxData => _outboxData;
```
Stephek: The Outbox is nothing to do with the Account.  It should be specified by a separate Interface and the implementation moved to the base class. Add as tech debt in 454.

src/SFA.DAS.EAS.Portal/Database/Models/Account.cs
```diff
@@ -0,0 +1,116 @@
...
+        
+        [JsonProperty("created")]
+        public DateTime Created { get; private set; }
```
Stephek: I believe the Created/Updated and Deleted relate to the stored document rather than the account read model.  They should be implemented from a separate interface and moved to the base class. Add as tech debt in 454.

src/SFA.DAS.EAS.Portal/Database/Models/Account.cs
```diff
@@ -0,0 +1,116 @@
...
+
+        private Account(long accountId, DateTime created, string messageId)
+            : base(1)
+        {
+            Id = Guid.NewGuid();
+            AccountId = accountId;
+            // the created date originates from event publishers, so we make sure we store it as UTC
+            Created = DateTime.SpecifyKind(created, DateTimeKind.Utc);
+            
+            AddOutboxMessage(messageId, created);
```
Stephek: Not a concern of the account read model.  This processing should be moved outside of the read model. Add as tech debt in 454.

src/SFA.DAS.EAS.Portal/Database/Models/Account.cs
```diff
@@ -0,0 +1,116 @@
...
+        }
+
+        // ctor for when creating account doc from reserved funding
+        // do we accept event, seems a coupling too far!
+        public Account(long accountId, long accountLegalEntityId, string legalEntityName, Guid reservationId,
```
Stephek: Looks very RPC like.  Doesn't support easy changes. Better to pass in objects.  Or leave it to the handlers to update the read model based on the events received. Add as tech debt in 454.

src/SFA.DAS.EAS.Portal/Database/Models/Account.cs
```diff
@@ -0,0 +1,116 @@
...
+            string courseId, string courseName, DateTime startDate, DateTime endDate, DateTime created, string messageId)
+            : this(accountId, created, messageId)
+        {
+            AddReserveFunding(accountLegalEntityId, legalEntityName, reservationId, courseId, courseName, startDate, endDate);
```
Stephek: Move to the handlers otherwise this object which is just a read model will become a big ball of mud as more requirements are met. Add as tech debt in 454.

src/SFA.DAS.EAS.Portal.Client/Application/Queries/GetAccountQuery.cs
lazcool: I'm not sure what you mean by streaming it direct. Could you please explain, or provide a link for me to read?

src/SFA.DAS.EAS.Portal/Database/Models/Account.cs
```diff
@@ -0,0 +1,116 @@
...
+        }
+
+        [JsonConstructor]
+        private Account()
+        {
+        }
+
+        public void AddReserveFunding(long accountLegalEntityId, string legalEntityName,  Guid reservationId,
```
Stephek: Better to move to the handler. Add as tech debt in 454.

src/SFA.DAS.EAS.Portal/Database/Models/Account.cs
```diff
@@ -0,0 +1,116 @@
...
+            });
+        }
+
+        private void AddReserveFunding(long accountLegalEntityId, string legalEntityName,
+            Guid reservationId, string courseId, string courseName, DateTime startDate, DateTime endDate)
+        {
+            _accountLegalEntities.Add(new AccountLegalEntity(accountLegalEntityId, legalEntityName, reservationId, courseId, courseName, startDate, endDate));
+        }
+
+        private void AddOutboxMessage(string messageId, DateTime created)
```
Stephek: This and following 3 methods are not a concern of the Account read model.  Needs moving out. Add as tech debt in 454.

src/SFA.DAS.EAS.Portal/Database/Models/Account.cs
```diff
@@ -0,0 +1,116 @@
...
+                
+                Updated = updated;
+                Deleted = null;
+            });
+        }
+
+        private void AddReserveFunding(long accountLegalEntityId, string legalEntityName,
+            Guid reservationId, string courseId, string courseName, DateTime startDate, DateTime endDate)
+        {
+            _accountLegalEntities.Add(new AccountLegalEntity(accountLegalEntityId, legalEntityName, reservationId, courseId, courseName, startDate, endDate));
```
Stephek: Doesn't prevent adding a reservation with the same Id twice.

src/SFA.DAS.EAS.Portal/Database/Models/AccountLegalEntity.cs
```diff
@@ -0,0 +1,46 @@
+using System;
+using System.Collections.Generic;
+using Newtonsoft.Json;
+using SFA.DAS.EAS.Portal.Client.Models;
+
+namespace SFA.DAS.EAS.Portal.Database.Models
+{
+    public class AccountLegalEntity : IAccountLegalEntityDto
```
Stephek: Port to Ashley's account read model. Add as tech debt in 454.

src/SFA.DAS.EAS.Portal/Database/Models/Document.cs
```diff
@@ -0,0 +1,30 @@
...
+
+        [JsonProperty("_etag")]
+        private string ReadOnlyETag { set => ETag = value; }
+
+        [JsonProperty("metadata")]
+        public DocumentMetadata Metadata { get; protected set;}
+
+        protected Document(short schemaVersion)
+        {
+            Metadata = new DocumentMetadata(schemaVersion);
```
Stephek: Good stuff.

src/SFA.DAS.EAS.Portal/Extensions/StringExtensions.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System.Net;
+
+namespace SFA.DAS.EAS.Portal.Extensions
+{
+    public static class StringExtensions
+    {
+        public static string HtmlDecode(this string input)
```
Stephek: Doesn't the SFA.DAS.NServiceBus.EndpointConfigurationExtensions handle the encoding of the license?  Maybe it should.

src/SFA.DAS.EmployerAccounts.Web/Controllers/EmployerTeamController.cs
```diff
@@ -348,9 +384,29 @@ public ActionResult CheckFunding(AccountDashboardViewModel model)
...
+        }
+        [ChildActionOnly]
         public ActionResult CreateVacancy(AccountDashboardViewModel model)
         {
             return PartialView(model);
         }
+
+        private async Task<OrchestratorResponse<AccountDashboardViewModel>> GetAccountInformation(string hashedAccountId)
+        {
+            var externalUserId = OwinWrapper.GetClaimValue(ControllerConstants.UserRefClaimKeyName);
```
Stephek: Should this logic be in the controller?

src/SFA.DAS.EAS.Portal.Client/Data/DocumentClientFactory.cs
lazcool: Do these values _need_ to be configurable? If we can come up with sensible values that are unlikely to change, keeping the number of config values down leads to simpler configuration. These values are the same as used by existing DAS Cosmos DB clients, which haven't had to be changed so far.

If we decide to make them configurable, I would suggest we allow the consumer of the client to supply the values, which they can either hard-code or read from their own config as appropriate.

src/SFA.DAS.EmployerAccounts.Web/Extensions/UrlHelperExtensions.cs
```diff
@@ -22,6 +22,14 @@ public static string EmployerCommitmentsAction(this UrlHelper helper, string pat
             return AccountAction(helper, baseUrl, path);
         }
 
+        public static string ReservationsAction(this UrlHelper helper, string path)
+        {
+            var configuration = DependencyResolver.Current.GetService<EmployerAccountsConfiguration>();
```
Stephek: Ouch.  Recommend against the service locator can the config not be passed in as an argument?

src/SFA.DAS.EmployerAccounts.Web/Views/EmployerTeam/V2/FundingComplete.cshtml
```diff
@@ -0,0 +1,22 @@
...
+        <dt class="das-definition-list__title">Legal entity:</dt>
+        <dd class="das-definition-list__definition">@Model.ReservedFundingToShowLegalEntityName</dd>
+        <dt class="das-definition-list__title">Training course:</dt>
+        <dd class="das-definition-list__definition">@Model.ReservedFundingToShow.CourseName</dd>
+        <dt class="das-definition-list__title">Start and end date: </dt>
+        <dd class="das-definition-list__definition">@Model.ReservedFundingToShow.StartDate.ToString("MMMM yyyy") to @Model.ReservedFundingToShow.EndDate.ToString("MMMM yyyy")</dd>
+    </dl>
+}
+
+@if (Model.RecentlyAddedReservationId != null &&
```
Stephek: Can this logic be put in the view model so we just use a bool here (e.g. Model.PendingFunding)

src/SFA.DAS.EmployerAccounts.Web/Views/EmployerTeam/V2/FundingComplete.cshtml
```diff
@@ -0,0 +1,22 @@
+﻿@model SFA.DAS.EmployerAccounts.Web.ViewModels.AccountDashboardViewModel
+
+@if (Model.ReservedFundingToShow != null)
```
Stephek: Please avoid the 'if' statement in the view if possible.  May be a case for two partials here.  One for when waiting and the other when the event has been received.  The waiting one could be reused elsewhere, the logic on what partial to use should be in the controller.

src/SFA.DAS.EAS.Portal/Database/Models/Account.cs
cofaulco: Worth noting that this has been raised and fixed as part of this branch, the Cosmos Db has a unique constraint against the AccountId, so the subsequent call to add would fail, retry the message and successfully get the account for the AccountId and add a new reservation.

src/SFA.DAS.EmployerAccounts.Web/Controllers/EmployerTeamController.cs
```diff
@@ -291,7 +299,35 @@ public ActionResult Row1Panel1(AccountDashboardViewModel model)
             {
                 viewModel.ViewName = "SignAgreement";
             }
+            else if (model.RecentlyAddedReservationId != null
+                || model.AccountViewModel?.AccountLegalEntities?.FirstOrDefault()?.ReservedFundings?.Any() == true)
```
Stephek: This logic should be in the view model.

src/SFA.DAS.EmployerAccounts.Web/Controllers/EmployerTeamController.cs
```diff
@@ -291,7 +299,35 @@ public ActionResult Row1Panel1(AccountDashboardViewModel model)
...
+                if (model.RecentlyAddedReservationId != null)
+                {
+                    var legalEntity = model.AccountViewModel?.AccountLegalEntities
+                        ?.FirstOrDefault(ale => ale.ReservedFundings?.Any(rf => rf.ReservationId == model.RecentlyAddedReservationId) == true);
+
+                    model.ReservedFundingToShowLegalEntityName = legalEntity?.LegalEntityName;
+
+                    // would be better to create new model to contain what the panel needs to show,
+                    // but we'll be replacing this with displaying all reserved funds anyway
+                    model.ReservedFundingToShow =
```
Stephek: Add this as a bool to the viewmodel

PR #1621, opened 2019-05-08
[CON-477] Add link to reservations to Check and secure funding

cofaulco:


src/SFA.DAS.EmployerAccounts.Web/Extensions/UrlHelperExtensions.cs
```diff
@@ -22,6 +22,14 @@ public static string EmployerCommitmentsAction(this UrlHelper helper, string pat
             return AccountAction(helper, baseUrl, path);
         }
 
+        public static string ReservationsAction(this UrlHelper helper, string path)
+        {
+            var configuration = DependencyResolver.Current.GetService<EmployerAccountsConfiguration>();
```
Stephek: Might be better to have this as a singleton and set it during the container creation.  Then a test could be written without having to use the DependencyResolver.  As this is the existing pattern I will approve it.

PR #1624, opened 2019-05-09
AML-3550 AML-3552 AML-3528 AML-3553

Howesy77:


src/SFA.DAS.EmployerAccounts.Web.UnitTests/Orchestrators/EmployerAccountOrchestratorTests/WhenCreatingTheUserAccount.cs
```diff
@@ -0,0 +1,92 @@
...
+
+            _mediator.Setup(x => x.SendAsync(It.IsAny<CreateUserAccountCommand>()))
+                .ReturnsAsync(new CreateUserAccountCommandResponse()
+                {
+                    HashedAccountId = "ABS10"
+                });
+        }
+
+        [Test]
+        public async Task ThenTheMediatorCommandIsCalledWithCorrectParameters()
```
MichaelYoung1981: Can we make the test names less technical?  They're describing the implementation which isn't ideal.

(I realise you copied these from somewhere else! :))

src/SFA.DAS.EmployerAccounts.Web/Controllers/EmployerAccountController.cs
```diff
@@ -136,25 +136,47 @@ public async Task<ActionResult> GateWayResponse()
...
+        [ValidateAntiForgeryToken]
+        [Route("getGovernmentFunding")]
+        public async Task<ActionResult> GetGovernmentFunding(int? choice)
+        {
+            switch (choice ?? 0)
             {
-                ViewBag.RequiresPayeScheme = string.IsNullOrEmpty(cookie.EmployerAccountPayeRefData.PayeReference) ||
-                                             cookie.EmployerAccountPayeRefData.EmpRefNotFound;
-            }
+                case 1:
```
MichaelYoung1981: Can the numbers be changed to an enum or constants so it's obvious what they're for?

src/SFA.DAS.EmployerAccounts/Commands/CreateUserAccount/CreateUserAccountCommandHandler.cs
```diff
@@ -0,0 +1,167 @@
...
+using SFA.DAS.HashingService;
+using SFA.DAS.NServiceBus;
+using SFA.DAS.Validation;
+using Entity = SFA.DAS.Audit.Types.Entity;
+
+
+namespace SFA.DAS.EmployerAccounts.Commands.CreateUserAccount
+{
+    //TODO this needs changing to be a facade and calling individual commands for each component
+    public class CreateUserAccountCommandHandler : IAsyncRequestHandler<CreateUserAccountCommand, CreateUserAccountCommandResponse>
```
MichaelYoung1981: There is something I don't like about this class and it needs a significant refactor (it's doing way too much).  I realise it was a copy from somewhere else and now probably isn't the appropriate time to resolve it but it's worth having a chat about on Monday maybe?

Winsto: @MichaelYoung1981  Dan's  //TODO comment at the top of the originating class is from early 2017, so definitely overdue a refactor :0)

PR #1636, opened 2019-05-16
Aml 3589 remove new journey feature toggle

Howesy77:


src/SFA.DAS.EmployerAccounts.Web.UnitTests/Controllers/EmployerAccountControllerTests/CreateAccount/Given_Cookie_Data_Is_Null/WhenICreateAnAccount.cs
```diff
@@ -36,8 +35,7 @@ public void Arrange()
 
             _orchestrator = new Mock<EmployerAccountOrchestrator>();
 
-            _owinWrapper = new Mock<IAuthenticationService>();
-            new Mock<IAuthorizationService>();
+            _owinWrapper = new Mock<IAuthenticationService>();
```
MichaelYoung1981: The namespace of these tests no longer makes sense, the Given_New_Journey... directory can be removed now.

Howesy77: It has.. I forgot to change the Namespace name

Winsto: Feel free to give those two scenarios more meaningful names, whilst you're in there if you feel like it :0)

Something like `Given_We_Have_Account_Data` and `Given_We_Do_Not_Have_Account_Data` would be better. It's only annoying me, because I wrote them ;0)

src/SFA.DAS/Authorization/FeatureType.cs
```diff
@@ -1,19 +1,19 @@
...
+    public enum FeatureType
+    {
+        NotSpecified,
+        Activities,
+        Projections,
+        ProviderRelationships,
+        Recruitments,
+        TransferConnectionRequests,
+        Transfers,
+        EnableNewRegistrationJourney,
```
Winsto: Are we keeping the feature enum value in if it has no purpose going forward?

Winsto:
`WhenISkipRegistration` test still has a reference to feature toggle; mocking it as true

https://github.com/SkillsFundingAgency/das-employerapprenticeshipsservice/blob/AML-3589_Remove_new_journey_feature_toggle/src/SFA.DAS.EmployerAccounts.Web.UnitTests/Controllers/EmployerAccountControllerTests/WhenISkipRegistration.cs

src/SFA.DAS.EmployerAccounts.Web.UnitTests/Controllers/EmployerAccountControllerTests/CreateAccount/Given_Cookie_Data_Is_Null/WhenICreateAnAccount.cs
```diff
@@ -81,8 +79,7 @@ public void Arrange()
                 _userViewTestingService.Object,
                 logger.Object,
                 _flashMessage.Object,
```
Winsto: There is still a reference to the feature toggle in the mocked authorization service; that can be completely removed now as it was only necessary for the feature toggle on and off scenarios.

src/SFA.DAS/Authorization/FeatureType.cs
Howesy77: Doesn't run if it's removed.

src/SFA.DAS.EmployerAccounts.Web.UnitTests/Controllers/EmployerAccountControllerTests/CreateAccount/Given_Cookie_Data_Is_Null/WhenICreateAnAccount.cs
Howesy77: Removed it

Winsto:
I have just realised we need : -

https://github.com/SkillsFundingAgency/das-employerapprenticeshipsservice/pull/1636

to go into master prior to this one; my comments regarding the unit test structure were on the assumption the change above had gone into master.

PR #1639, opened 2019-05-21
Con 436 add paye and orgs

cofaulco:


src/SFA.DAS.EmployerAccounts.Web/Controllers/EmployerTeamController.cs
```diff
@@ -50,16 +52,19 @@ public class EmployerTeamController : BaseController
         public async Task<ActionResult> Index(string hashedAccountId, string reservationId)
         {
             var response = await GetAccountInformation(hashedAccountId);
-            if (FeatureToggles.Features.HomePage.Enabled)
+            if (FeatureToggles.Features.HomePage.Enabled || response.Data.PayeSchemeCount == 0)
```
Stephek: Might be cleaner to extract in to a method like HasPayeScheme(Account).

src/SFA.DAS.EmployerAccounts.Web/Controllers/EmployerTeamController.cs
```diff
@@ -294,48 +299,15 @@ public async Task<ActionResult> HideWizard(string hashedAccountId)
...
             return PartialView(viewModel);
-        }       
+        }
 
         [ChildActionOnly]
         public ActionResult Row1Panel2(AccountDashboardViewModel model)
         {
             var viewModel = new PanelViewModel<AccountDashboardViewModel> { ViewName = "ProviderPermissions", Data = model };
-            if (model.AgreementsToSign)
+            if (model.PayeSchemeCount == 0 || model.AgreementsToSign)
```
Stephek: This should be a single flag/property on the model so the code here is something like

if(model.CanAddProviderPermissions)
{
}

src/SFA.DAS.EmployerAccounts.Web/Helpers/HomepagePanelViewHelper.cs
```diff
@@ -0,0 +1,59 @@
+﻿using System.Linq;
+using SFA.DAS.EmployerAccounts.Web.ViewModels;
+
+namespace SFA.DAS.EmployerAccounts.Web.Helpers
+{
+    public class NextActionPanelViewHelper : INextActionPanelViewHelper
```
Stephek: This is too specific to PAYE.  I think the panel name should be passed in so it can be used elsewhere.  Also rather than return a AccountDashboardViewModel return a NextView (?} object or ViewInfo object with a name property that has the name of the view to render next.  That way this helpers concern is just to give information about the next view to the caller rather than tying it to a specific viewmodel.

cofaulco: I agree, however I think once this is merged with 454 there will be a large refactor here anyway so this code will change anyway

PR #1632, opened 2019-05-14
Con 454 homepage panels & models

Ashley-Bass:
Incudes:
CON-393
CON-414
~~CON-451~~ Already in master

PR #1646, opened 2019-05-24
CON 409 Provider Permissions Panel

lazcool:
[Story](https://skillsfundingagency.atlassian.net/browse/CON-409)

Add subscription to `AddedAccountProviderEvent` to capture when a provider relationship is created, then saves the provider's details in local store.

Displays single provider details in a panel on the portal.

Also has CON-503 merged in, which handles the panel when there are multiple providers, and a fix on top of CON-503 to make it work.

Bonuses:
* Fixes bug where cohort is added to incorrect organisation when processing `CohortApprovalRequestedByProvider` event.
* Add missing unit tests around `ReservationCreatedEvent` handling.
* Refactoring.


src/SFA.DAS.EAS.Portal.Client/Database/Models/AccountDocument.cs
```diff
@@ -11,20 +11,16 @@ public class AccountDocument : Document
 
         [JsonProperty("accountId")]
         public long AccountId => Account.Id;
-        private AccountDocument(long accountId) : base(1)
+        public AccountDocument(long accountId) : base(1)
```
Stephek: Not sure why this has changed In this story but the factory pattern may be preferable to protect from future changes/creation rules.  Raise as a tech debt issue if you think it should be changed.  Please stick to the story/use case.

lazcool: If there a need for a factory method at a later date, we can introduce one. At the moment, I can't see any value in having one.

src/SFA.DAS.EAS.Portal.UnitTests/Portal/Application/Commands/Cohort/CohortApprovalRequestedCommandTests.cs
```diff
@@ -58,52 +61,26 @@ public TestContext()
...
             [Test]
             public async Task WhenCalled_ThenTheAccountServiceIsCalledToSaveTheAccount()
             {
                 // arrange
                 var testContext = new TestContext();
-                CohortApprovalRequestedCommand command = new CohortApprovalRequestedCommandBuilder();
 
                 // act
-                await testContext.Sut.Handle(command);
+                await testContext.CohortApprovalRequestedCommand.Execute(testContext.CohortApprovalRequestedByProvider);
```
Stephek: What has this got to do with the story?  Please raise as tech debt for discussion if you think things need to change.

src/SFA.DAS.EAS.Portal.UnitTests/Portal/Application/Commands/Cohort/CohortApprovalRequestedCommandTests.cs
```diff
@@ -140,19 +115,16 @@ public async Task WhenCalledForAnExistingCohort_ThenTheCohortCountIsNotChanged()
                 testContext.TestCommitment.Id = cohortId;
                 testContext.TestAccount.Organisations.First().Cohorts.Count.Should().Be(1);
 
-                CohortApprovalRequestedCommand command = new CohortApprovalRequestedCommandBuilder();
-
                 // act
-                await testContext.Sut.Handle(command);
+                await testContext.CohortApprovalRequestedCommand.Execute(testContext.CohortApprovalRequestedByProvider);
```
Stephek: Not part of this story.

src/SFA.DAS.EAS.Portal.UnitTests/Portal/Application/Commands/Cohort/CohortApprovalRequestedCommandTests.cs
```diff
@@ -164,13 +136,11 @@ public async Task WhenANewCohortIsAdded_ThenApprenticeshipsInTheCohortAreStored(
                 testContext.TestCommitment.Reference = cohortReference;
                 testContext.TestCommitment.Apprenticeships = new List<Commitments.Api.Types.Apprenticeship.Apprenticeship>() { new Commitments.Api.Types.Apprenticeship.Apprenticeship { Id = apprenticeshipId } };
 
-                CohortApprovalRequestedCommand command = new CohortApprovalRequestedCommandBuilder();
-
                 // act
-                await testContext.Sut.Handle(command);
+                await testContext.CohortApprovalRequestedCommand.Execute(testContext.CohortApprovalRequestedByProvider);
```
Stephek: Not part of this story.

src/SFA.DAS.EAS.Portal.UnitTests/Portal/Application/Commands/Cohort/CohortApprovalRequestedCommandTests.cs
```diff
@@ -206,13 +176,11 @@ public async Task WhenAnExistingCohortIsHandled_ThenAnyApprenticeshipsChangesAre
 
                 testContext.TestAccount.Organisations.First().Cohorts.First().Apprenticeships.Count.Should().Be(1);
 
-                CohortApprovalRequestedCommand command = new CohortApprovalRequestedCommandBuilder();
-
                 // act
-                await testContext.Sut.Handle(command);
+                await testContext.CohortApprovalRequestedCommand.Execute(testContext.CohortApprovalRequestedByProvider);
```
Stephek: Not part of this story.

src/SFA.DAS.EAS.Portal.Client/Database/Models/Document.cs
```diff
@@ -12,7 +12,7 @@ public abstract class Document : IDocument, IDocumentProperties, IOutbox
         public Guid Id { get; protected set; }
 
         [JsonIgnore]
-        public bool IsNew { get; protected set; }
+        public bool IsNew { get; set; }
```
Stephek: This is an internal property and shouldn't be able to be set from outside the object.  Please put back to protected.

src/SFA.DAS.EAS.Portal.UnitTests/Portal/Application/Commands/Cohort/CohortApprovalRequestedCommandTests.cs
lazcool: Ideally yes, each tech debt item should have a separate PR.

However, in the initial creation phase of a project, when deadlines are so tight, it's useful to relax the overhead to get the system built in time, as you found useful in the [454 PR](https://github.com/SkillsFundingAgency/das-employerapprenticeshipsservice/pull/1632), which also contained many refactors 😉.

We should be in a position now though, going forward, to have separate PR's.

src/SFA.DAS.EAS.Portal.Client/Database/Models/Document.cs
lazcool: I agree it shouldn't be public, it was a (pained 😉) tactical decision to make it public, so that we could have a working unit test, without extending the time taken to complete the story. If `Document` was part of the public API, I'd have addressed it before declaring the PR ready to review, but I didn't, as it's only used internally in the Portal (the client only returns the contained `Account` object).

There are also other properties in `Document` which look like they should have more restrictive access modifiers. However, to make them more restrictive, we'd have to either refactor the code so that the test code doesn't need to set the property, or introduce a separate mechanism to allow the tests to modify it. (I know ideally tests shouldn't need to modify non-public properties!)

We could declare it as `protected internal` and make internals visible to the unit test project.

I can raise a tech debt item to address the access modifiers, unless you think that we need to address it as part of this PR?

src/SFA.DAS.EAS.Portal.Worker/EventHandlers/EventHandler.cs
```diff
@@ -0,0 +1,25 @@
+using System.Threading.Tasks;
+using NServiceBus;
+using SFA.DAS.EAS.Portal.Application.Commands;
+using SFA.DAS.EAS.Portal.Application.Services;
+
+namespace SFA.DAS.EAS.Portal.Worker.EventHandlers
+{
+    public class EventHandler<TEvent> : IHandleMessages<TEvent>
```
Stephek: Nice one.

src/SFA.DAS.EAS.Portal/Application/Commands/Command.cs
```diff
@@ -0,0 +1,72 @@
...
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+using SFA.DAS.EAS.Portal.Application.Services;
+using SFA.DAS.EAS.Portal.Client.Database.Models;
+using SFA.DAS.EAS.Portal.Client.Types;
+
+namespace SFA.DAS.EAS.Portal.Application.Commands
+{
+    public class Command
```
Stephek: This is poor design and will lead to spaghetti code.  We shouldn't have a base command trying to implement all this functionality e.g. GetOrAddProvider, GetOrAddCohort.  That is not the concern of a command.  A command is a lightweight object containing some data.

src/SFA.DAS.EAS.Portal/Application/Commands/Cohort/CohortApprovalRequestedCommand.cs
```diff
@@ -1,16 +1,72 @@
...
+using SFA.DAS.HashingService;
+
+namespace SFA.DAS.EAS.Portal.Application.Commands.Cohort
 {
-    public class CohortApprovalRequestedCommand : ICommand
-    {        
-        public long AccountId { get; private set; }
-        public long ProviderId { get; private set; }
-        public long CommitmentId { get; private set; }
+    public class CohortApprovalRequestedCommand : Command, ICommand<CohortApprovalRequestedByProvider>
```
Stephek: We should separate the concerns of the command from the handler for that command.  Is this a command or a handler?  Stick to the single responsibility principle to keep the code clean easy to test and easy to change.

src/SFA.DAS.EAS.Portal.Worker/EventHandlers/EventHandler.cs
lazcool: Thanks 👍

src/SFA.DAS.EAS.Portal/Application/Services/IMessageContext.cs
```diff
@@ -1,10 +1,13 @@
...
+using NServiceBus;
 
 namespace SFA.DAS.EAS.Portal.Application.Services
 {
     public interface IMessageContext
     {
         string Id { get; set; }
         DateTime CreatedDateTime { get; set; }
+
+        void Initialise(IMessageHandlerContext handlerContext);
```
Stephek: It doesn't make sense for Intialise with a IMessageHandlerContext to be on this interface.  I understand why you have (to remove the extension method).  To achieve this it would be better design to create another interface for this method that MessageContext implements.  This interface design is intended for the consumers later in the flow they have no dependency on IMessageHandlerContext.

src/SFA.DAS.EAS.Portal/DependencyResolution/ApplicationServices.cs
```diff
@@ -1,43 +1,23 @@
...
             services.AddTransient<IAccountDocumentService, AccountDocumentService>();                        
             services.Decorate<IAccountDocumentService, AccountDocumentServiceWithSetProperties>();
             services.Decorate<IAccountDocumentService, AccountDocumentServiceWithDuplicateCheck>();
-            services.AddScoped<IAccountHelperService, AccountHelperService>();
-            services.AddTransient<IAdapter<CohortApprovalRequestedByProvider, CohortApprovalRequestedCommand>, CohortAdapter>();
 
-            // Register all ICommandHandler<> types
             services.Scan(scan =>
-                    scan.FromAssembliesOf(typeof(ICommandHandler<>))
+                    scan.FromAssembliesOf(typeof(ICommand<>))
```
Stephek: If it looks like a Handler and walks like a Handler it is a Handler.  Please rename back to ICommandHandler.

src/SFA.DAS.EAS.Portal/DependencyResolution/HashingServices.cs
```diff
@@ -0,0 +1,20 @@
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using SFA.DAS.EAS.Portal.Configuration;
+using SFA.DAS.HashingService;
+
+namespace SFA.DAS.EAS.Portal.DependencyResolution
+{
+    public static class HashingServices
```
Stephek: Good tidy up.

src/SFA.DAS.EAS.Portal/DependencyResolution/ProviderServices.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace SFA.DAS.EAS.Portal.DependencyResolution
+{
+    public static class ProviderServices
+    {
+        public static IServiceCollection AddProviderServices(this IServiceCollection services, HostBuilderContext hostBuilderContext)
+        {
+            var apiClientConfig = hostBuilderContext.Configuration.GetSection<ApprenticeshipInfoServiceApiConfiguration>(ConfigurationKeys.ApprenticeshipInfoServiceApi);
+
+            services.AddTransient<IProviderApiClient>(sp => new ProviderApiClient(apiClientConfig.BaseUrl));
+            // if we want to support config changes on the fly, we could create wrapper class that accepts IOptionsMonitor and picks out url
```
Stephek: Do we need these comments?  Create a tech debt item instead rather than litter the code with your thoughts.

src/SFA.DAS.EmployerAccounts.Web/Controllers/EmployerTeamController.cs
```diff
@@ -332,11 +332,20 @@ public ActionResult Row1Panel1(AccountDashboardViewModel model)
         [ChildActionOnly]
         public ActionResult Row1Panel2(AccountDashboardViewModel model)
         {
-            var viewModel = new PanelViewModel<AccountDashboardViewModel> { ViewName = "ProviderPermissions", Data = model };
+            var viewModel = new PanelViewModel<AccountDashboardViewModel> { Data = model };
             if (model.PayeSchemeCount == 0 || model.AgreementsToSign)
             {
                 viewModel.ViewName = "ProviderPermissionsDenied";
             }
+            else if (model.AccountViewModel?.Providers?.Any() ?? false)
```
Stephek: The model should provide a flag to determine the state to display.  Please move this logic to the model.

src/SFA.DAS.EAS.Portal/Application/Commands/Command.cs
lazcool: We had this discussion in the 454 PR 😄! Here's my position from one of the relevant [comments ](https://github.com/SkillsFundingAgency/das-employerapprenticeshipsservice/pull/1632#discussion_r284327408)from that PR, so that we can continue the discussion...

> Stephek 20 days ago  Member
> Is this a command or command handler? Should be separation of concerns as per the standard pattern command.handler pattern.
> 
> lazcool 19 days ago  Member
> It's a command object and there is no 'standard' pattern.
> 
> From [wikipedia's page on the command pattern](https://en.wikipedia.org/wiki/Command_pattern):
> 'In some implementations the handler code is part of the command/action object. In other implementations the code is part of the Receiver/Target Object, and in yet other implementations the handler code is kept separate from the other objects.'
> 

We also discussed the Command vs Command+Command Handler pattern in another [comment](https://github.com/SkillsFundingAgency/das-employerapprenticeshipsservice/pull/1632#discussion_r286857670). My point from that comment was, essentially, that the pattern where we have just a Command class is simpler, requires less code and performs better.

If we accept that a command object can encapsulate a command on its own, then it makes sense to put common functionality in a base class.

src/SFA.DAS.EAS.Portal/Application/Commands/Cohort/CohortApprovalRequestedCommand.cs
lazcool: This is essentially the same point as your previous [comment](https://github.com/SkillsFundingAgency/das-employerapprenticeshipsservice/pull/1646#discussion_r290274089), which I responded to in that thread.

My personal opinion is that having a single command means cleaner code, and I find it easy to test and change!

src/SFA.DAS.EAS.Portal/DependencyResolution/ApplicationServices.cs
lazcool: Funny! 😂

If it was an interface for a command handler, then it would be appropriate to call the interface `ICommandHandler`. As it's an interface for a command, then `ICommand` is appropriate.

Once we decide on the Command vs Command+Command Handler pattern, we can name the interfaces appropriately.

src/SFA.DAS.EAS.Portal.Client/Database/Models/AccountDocument.cs
Stephek: To encapsulate the setting of the IsNew flag.

src/SFA.DAS.EAS.Portal.Client/Database/Models/Document.cs
Stephek: A unit test shouldn't have to set it.

src/SFA.DAS.EAS.Portal/DependencyResolution/HashingServices.cs
lazcool: Thanks! It was a delight to find the `HostBuilderContext`, as it solves a lot of messy workarounds.

src/SFA.DAS.EAS.Portal/Application/Commands/Command.cs
Stephek: Base classes should be for inheritance not common code.

src/SFA.DAS.EAS.Portal/DependencyResolution/ProviderServices.cs
lazcool: Litter removed! :boom:

lazcool:
> The main issues here are making changes unrelated to the requirements. This has delayed the release of this story to test. Apart from that changing the IsNew property on the Document class needs to be reverted and the Command and handler responsibilities should be separated. We should avoid base classes unless there is a genuine inheritance hierarchy, they are not the place for common code therefore there is no justification for a base command class.

All the issues you've raised in this comment are being addressed individually in the threads from your original comments. The only new issue raised in this comment, is with regard to there being a command base class, so I'll address that here.

From Microsoft's [document on inheritance](https://docs.microsoft.com/en-us/dotnet/csharp/tutorials/inheritance)...

> Ordinarily, inheritance is used to express an "is a" relationship between a base class and one or more derived classes, where the derived classes are specialized versions of the base class; the derived class is a type of the base class.

In my opinion, a `AddAccountProviderCommand` "is a" command. (In addition I also think a `AddReservationCommand` "is a" command, and a `CohortApprovalRequestedCommand` "is a" command.

Also, opinion in the answers to this [stack overflow question](https://stackoverflow.com/questions/32831991/is-it-good-practice-to-move-common-code-to-base-class-assuming-there-is-none) seems to be aligned... 

> Is it good practice to move common code to base class [...]? 
> 
> * Yes, absolutely...
> * It's really a good idea...
> * This seems like a perfectly reasonable solution...
> 

Although saying all that, I do generally agree with the maxim 'prefer composition over inheritance' :smile:, but from the book Code Complete (referenced on this [stack overflow page](https://stackoverflow.com/questions/49002/prefer-composition-over-inheritance))... 

> Think of containment as a has a relationship. A car "has an" engine, a person "has a" name, etc.
> Think of inheritance as an is a relationship. A car "is a" vehicle, a person "is a" mammal, etc.

So in this particular case, I think `AddAccountProviderCommand` "is a" command, is more appropriate than `AddAccountProviderCommand` "has a" command.

Another way of thinking about 'Should I be inheriting from this type?' is using 'Liskov Substitution Principle' as a test. From the [Wikipedia article on LSP](https://en.wikipedia.org/wiki/Liskov_substitution_principle)...

> ...if S is a subtype of T, then objects of type T in a program may be replaced with objects of type S without altering any of the desirable properties of that program (e.g. correctness).

In this particular instance, we don't currently have anything that accepts just a `Command`, but we could. E.g. we could have a common test fixture for the command unit tests that accepts a command, and it would be valid to call the test with one of the derived commands.

lazcool:
> > The main issues here are making changes unrelated to the requirements. This has delayed the release of this story to test. Apart from that changing the IsNew property on the Document class needs to be reverted and the Command and handler responsibilities should be separated. We should avoid base classes unless there is a genuine inheritance hierarchy, they are not the place for common code therefore there is no justification for a base command class.
> 
> All the issues you've raised in this comment are being addressed individually in the threads from your original comments. The only new issue raised in this comment, is with regard to there being a command base class, so I'll address that here.
> 
> From Microsoft's [document on inheritance](https://docs.microsoft.com/en-us/dotnet/csharp/tutorials/inheritance)...
> 
> > Ordinarily, inheritance is used to express an "is a" relationship between a base class and one or more derived classes, where the derived classes are specialized versions of the base class; the derived class is a type of the base class.
> 
> In my opinion, a `AddAccountProviderCommand` "is a" command. (In addition I also think a `AddReservationCommand` "is a" command, and a `CohortApprovalRequestedCommand` "is a" command.
> 
> Also, opinion in the answers to this [stack overflow question](https://stackoverflow.com/questions/32831991/is-it-good-practice-to-move-common-code-to-base-class-assuming-there-is-none) seems to be aligned...
> 
> > Is it good practice to move common code to base class [...]?
> > 
> > * Yes, absolutely...
> > * It's really a good idea...
> > * This seems like a perfectly reasonable solution...
> 
> Although saying all that, I do generally agree with the maxim 'prefer composition over inheritance' 😄, but from the book Code Complete (referenced on this [stack overflow page](https://stackoverflow.com/questions/49002/prefer-composition-over-inheritance))...
> 
> > Think of containment as a has a relationship. A car "has an" engine, a person "has a" name, etc.
> > Think of inheritance as an is a relationship. A car "is a" vehicle, a person "is a" mammal, etc.
> 
> So in this particular case, I think `AddAccountProviderCommand` "is a" command, is more appropriate than `AddAccountProviderCommand` "has a" command.
> 
> Another way of thinking about 'Should I be inheriting from this type?' is using 'Liskov Substitution Principle' as a test. From the [Wikipedia article on LSP](https://en.wikipedia.org/wiki/Liskov_substitution_principle)...
> 
> > ...if S is a subtype of T, then objects of type T in a program may be replaced with objects of type S without altering any of the desirable properties of that program (e.g. correctness).
> 
> In this particular instance, we don't currently have anything that accepts just a `Command`, but we could. E.g. we could have a common test fixture for the command unit tests that accepts a command, and it would be valid to call the test with one of the derived commands.

Sorry, the issue of using a base class was raised in an existing comment!

src/SFA.DAS.EAS.Portal.Client/Database/Models/Document.cs
lazcool: `IsNew` has been changed back to having a protected set. The unit test now creates the `AccountDocument` by deserializing it from json, which has the added bonus of more accuratly mirroring how its created in the test scenario.

src/SFA.DAS.EAS.Portal.Client/Database/Models/AccountDocument.cs
lazcool: The `IsNew` flag is always set to `true` when constructed through the public constructor. (I've made the `JsonConstructor` attributed constructor private, as it provided a back door to constructing an `AccountDocument` without setting up invariants.)

The construction of `AccountDocument` is encapsulated in the constructor. We don't need to encapsulate setting individual properties that the constructor sets.

src/SFA.DAS.EAS.Portal/Application/Commands/Command.cs
lazcool: Please see my other [comment](https://github.com/SkillsFundingAgency/das-employerapprenticeshipsservice/pull/1646#issuecomment-498708165), which details how this adhers to the Liskov Substitution Principle (re your comment about following SOLID).

src/SFA.DAS.EAS.Portal/Application/Services/IMessageContext.cs
lazcool: Good point. I've taken `Initialise` out of `IMessageContext` and moved it into a new `IMessageContextInitialisation` interface.

src/SFA.DAS.EmployerAccounts.Web/Controllers/EmployerTeamController.cs
lazcool: The logic has been moved into the model. We can introduce a flag if the property is called more than once.

src/SFA.DAS.EAS.Portal/Application/Commands/Command.cs
Stephek: GetOrAddProvider, GetOrAddCohort is not common command code.  As per my original comment this code should not reside in a Command base object and therefore there is no need for a base object.

PR #1652, opened 2019-05-28
AML-3568: update contents as per story.

Winsto:


DannyGB:
Original author left and the PR is over 9 months old, it will never be merged

PR #1657, opened 2019-05-30
Rebase of account get or create to accountdocumentservice

Ashley-Bass:


src/SFA.DAS.EAS.Portal/Application/Services/AccountDocumentService.cs
```diff
@@ -14,12 +14,15 @@ public AccountDocumentService(IAccountsRepository accountsRepository)
         {
             _accountsRepository = accountsRepository;
         }
-        public Task<AccountDocument> Get(long id, CancellationToken cancellationToken = default)
+        public async Task<AccountDocument> Get(long id, CancellationToken cancellationToken = default)
```
cofaulco: Get but will create if not exists?

src/SFA.DAS.EAS.Portal.UnitTests/Portal/Application/Services/AccountDocumentServicesTest.cs
```diff
@@ -0,0 +1,84 @@
...
+using Microsoft.ApplicationInsights.Metrics.Extensibility;
+using Microsoft.Azure.Documents.Client;
+
+namespace SFA.DAS.EAS.Portal.UnitTests.Portal.Application.Services
+{
+    [Parallelizable]
+    [TestFixture]
+    class AccountDocumentServicesTest
+    {
+        public class TestContext
```
cofaulco: What purpose does the test context serve here? all this logic could be done it SetUp and therefore no need for the extra class with a bunch of public properties

Ashley-Bass: It's the same way the other tests are structured, so I just followed the pattern, happy to switch if preferred

cofaulco:
Old stale branch no longer required

PR #1659, opened 2019-05-31
AML-3578 Removed 'What You'll Need' page

Howesy77:


src/SFA.DAS.EAS.Web/Helpers/ControllerConstants.cs
```diff
@@ -6,44 +6,9 @@ public static class ControllerConstants
         public const string AccountHashedIdRouteKeyName = "HashedAccountId";
         public const string ActionKeyName = "Action";
         public const string ActivitiesUnavailableMessage = "Activities are currently unavailable.";
-        public const string AddOrganisationAddressViewName = "../OrganisationShared/AddOrganisationAddress";
```
Winsto: Where all of these rendered redundant by the removal of the "What y'all need" page? Or was there some legacy redundancy hanging around? Seems like a lot of constants removed for just one page.

Howesy77: Removed as unused so tidied up code.

PR #1660, opened 2019-05-31
Aml 3667 add paye and org to existing account

Winsto:
Proof of concept code done to provide quick fix to multi account issue.

Although a bit quick and crufty we had three devs working on this, so we can just have QA approve.

### DIFF
--- a/src/SFA.DAS.EmployerAccounts.Web/Models/CreateAccountModel.cs
+++ b/src/SFA.DAS.EmployerAccounts.Web/Models/CreateAccountModel.cs
@@ -1,10 +1,14 @@
 ﻿using System;
 using SFA.DAS.Common.Domain.Types;
 
-namespace SFA.DAS.EmployerAccounts.Web.ViewModels
+namespace SFA.DAS.EmployerAccounts.Web.Models
 {
-    public class CreateAccountViewModel
+    public class CreateAccountModel
     {
+        public CreateAccountModel()
+        {
+            HashedAccountId = new HashedAccountIdModel {Value = String.Empty};
+        }
         public string UserId { get; set; }
         public OrganisationType OrganisationType { get; set; }
         public short? PublicSectorDataSource { get; set; }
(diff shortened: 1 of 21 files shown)


src/SFA.DAS.EmployerAccounts.Web/Models/CreateAccountModel.cs
```diff
@@ -18,5 +18,6 @@ public class CreateAccountViewModel
         public string OrganisationStatus { get; set; }
         public string EmployerRefName { get; set; }
         public string Sector { get; set; }
+        public HashedAccountIdModel HashedAccountId { get; set; }
```
cofaulco: Is there a reason that this isn't just a string ?

src/SFA.DAS.EmployerAccounts.Web/Orchestrators/EmployerAccountOrchestrator.cs
```diff
@@ -117,25 +121,110 @@ public virtual async Task<OrchestratorResponse<RenameEmployerAccountViewModel>>
...
+                {
+                    Data = new EmployerAgreementViewModel(),
+                    Status = HttpStatusCode.BadRequest,
+                    Exception = e,
+                    FlashMessage = new FlashMessageViewModel()
+                };
+            }
+        }
+
+        private async Task updateAccountNameToLegalEntityName(CreateAccountModel model)
```
Howesy77: Function names should start Uppercase?

src/SFA.DAS.EmployerAccounts.Web/Models/CreateAccountModel.cs
Winsto: Err, just we already had a value type so kept it I think really, rather than go with primitive.

In this case the value type is merely a wrapper for a string  (no validation at construction etc. ), so wouldn't make much difference.

There could be an argument (in a re-factor) for this to not even be part of this model and the hashedaccountid storage retrieval to be 'behind' the Orchestrator.

src/SFA.DAS.EmployerAccounts.Web/Orchestrators/EmployerAccountOrchestrator.cs
Winsto: Updated.

PR #1661, opened 2019-06-03
AML-3522 Check your details page

Howesy77:
All looks ok, just table formatting issue.


Winsto:
Noting team discussion: the functionality requested in the story is existing functionality.

The code change here is for some tidy up that we may as well put in.

PR #1664, opened 2019-06-04
AML-3665 Content change for redesign

Howesy77:


src/SFA.DAS.EmployerAccounts.Web/Views/EmployerAccount/GetGovernmentFunding.cshtml
```diff
@@ -11,11 +11,15 @@
 
 <div class="grid-row">
     <div class="column-two-thirds">
-        <h1 class="heading-xlarge">
-            Get government funding
+        <h1 class="heading-xlarge">
+            Get government funding
```
tom-gough: what is the change here? looks identical

Howesy77: There isn't one.. Why has it highlighted that?

PR #1665, opened 2019-06-04
Change to start of journey

Howesy77:


src/SFA.DAS.EAS.Web/Controllers/HomeController.cs
```diff
@@ -32,13 +32,6 @@ public ActionResult ViewAccounts()
             return Redirect(Url.EmployerAccountsAction("service/accounts", false));
         }
 
-        [HttpGet]
-        [Route("usedServiceBefore")]
-        public ActionResult UsedServiceBefore()
-        {
-            return Redirect(Url.EmployerAccountsAction("service/usedServiceBefore", false));
-        }
```
tom-gough: Assume whatever points to this route is no longer doing so?

Howesy77: Yeah the start page goes straight to the sign-in or registration page.

PR #1669, opened 2019-06-06
Aml 3657 tidy up sfa.das.eas.application

Howesy77:


src/SFA.DAS.EAS.Api/Web.config
```diff
@@ -1,389 +1,389 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
```
Winsto: Did this file really change? I can't see any clear diffs; looks a bit like a complete delete and add.

Was it in commit by mistake? or is it merge noise?

src/SFA.DAS.EAS.Web.IntegrationTests/app.config
```diff
@@ -1,157 +1,157 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
```
Winsto: This looks like an incorrect commit.

PR #1674, opened 2019-06-07
Con 542 projected calculations

Stephek:


src/SFA.DAS.EmployerFinance/Services/DasForecastingService.cs
```diff
@@ -74,5 +75,57 @@ public async Task<ExpiringAccountFunds> GetExpiringAccountFunds(long accountId)
...
             }
         }
+
+        public async Task<ProjectedCalculation> GetProjectedCalculations(long accountId)
+        {
+            _logger.Info($"Getting projected calculations for account ID: {accountId}");
+
+            var expiredFundsUrl = $"/api/accounts/{accountId}/AccountProjection/projected-summary";
+
+            var accessToken = await _azureAdAuthService.GetAuthenticationResult(
```
cofaulco: Could we reuse common code between this method and the Expired funds one? They are using the same api

Stephek: Done

PR #1676, opened 2019-06-10
AML-3693 and code tidy up for other issues

Howesy77:


src/SFA.DAS.EmployerAccounts.Web/Views/Home/Index.cshtml
```diff
@@ -1,89 +1,78 @@
-@using SFA.DAS.Authorization
```
tom-gough: Has every line of this file legitimately changed?

PR #1677, opened 2019-06-10
Aml 3694 fix multi account creation when going back and forth

Winsto:
Two approaches in this history; first attempt failed, because the return type of a query lied about what it actually was :0(

This is a bit crufty, but works with local testing.

### DIFF
--- a/src/SFA.DAS.EmployerAccounts.Web/Orchestrators/EmployerAccountOrchestrator.cs
+++ b/src/SFA.DAS.EmployerAccounts.Web/Orchestrators/EmployerAccountOrchestrator.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Linq;
 using MediatR;
 using SFA.DAS.Authorization;
 using SFA.DAS.EmployerAccounts.Commands.RenameEmployerAccount;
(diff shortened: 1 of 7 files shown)


src/SFA.DAS.EmployerAccounts.Web/Orchestrators/EmployerAccountOrchestrator.cs
```diff
@@ -252,10 +255,23 @@ private async Task<OrchestratorResponse<EmployerAgreementViewModel>> CreateNewAc
...
 
-        public virtual async Task<OrchestratorResponse<EmployerAccountViewModel>> CreateUserAccount(CreateUserAccountViewModel viewModel, HttpContextBase context)
+        public virtual async Task<OrchestratorResponse<EmployerAccountViewModel>> CreateMinimalUserAccountForSkipJourney(CreateUserAccountViewModel viewModel, HttpContextBase context)
         {
             try
             {
+                var existingUserAccounts =
+                    await Mediator.SendAsync(new GetUserAccountsQuery {UserRef = viewModel.UserId});
+
+                if(existingUserAccounts?.Accounts?.AccountList?.Any() == true)
```
tom-gough: why do we need "== true" here?

tom-gough: ah because it could be null :)

PR #1683, opened 2019-06-11
Con 536 apprentice approved event

Ashley-Bass:


Ashley-Bass:
As current will need the commitments package with the new event, which then means we can create a new portal package and then update the frontend

src/SFA.DAS.EAS.Portal.UnitTests/Worker/EventHandlers/Commitments/CohortApprovedByEmployerEventHandlerTests.cs
```diff
@@ -0,0 +1,78 @@
...
+            AccountDocumentService.Verify(
+                s => s.Save(It.Is<AccountDocument>(d => CohortIsSetToApproved(d)),It.IsAny<CancellationToken>()), Times.Once);
+            
+            return this;
+        }
+        
+        private bool CohortIsSetToApproved(AccountDocument document)
+        {
+            return document.Account.Organisations.SelectMany(org => org.Cohorts)
+                .SingleOrDefault(co => co.Id == Message.CommitmentId.ToString()).IsApproved;
```
Stephek: Just use .Single(

Ashley-Bass: done

PR #1694, opened 2019-06-12
Aml 3697 aml 3698

Howesy77:


src/SFA.DAS.EmployerAccounts.Web/Orchestrators/SearchPensionRegulatorOrchestrator.cs
```diff
@@ -80,7 +78,8 @@ private PensionRegulatorDetailsViewModel ConvertToViewModel(Organisation organis
                 Name = organisation.Name,
                 Status = organisation.Status,
                 Address = organisation.Address.FormatAddress(),
-                ReferenceNumber = organisation.UniqueIdentity
+                ReferenceNumber = organisation.UniqueIdentity,
+                Type = OrganisationType.PensionsRegulator
```
MichaelYoung1981: Shouldn't this have resulted in a change to a test?

src/SFA.DAS.EmployerAccounts.Web/Views/EmployerAccount/Summary.cshtml
```diff
@@ -27,30 +27,30 @@
                     Address
                 </th>
                 <td colspan="2">
-                    @Model.Data.RegisteredAddress
+                    @foreach (var line in Model.Data.RegisteredAddress.Split(','))
```
MichaelYoung1981: Could we move the split into the model so it can be tested?

src/SFA.DAS.EmployerAccounts.Web/Orchestrators/SearchPensionRegulatorOrchestrator.cs
Winsto: I have added a test to cover this. However I don't really think this property should be being set at this 'high' level. I think really this 'source' property should be set way down at the point it is retrived in the `IPensionRegulatorService` implementation.

I had a quick look at doing that, but I think the best way would be to hook into the Json deserialization and didn't have a quick answer for that :0) 

We can sort that out as part of this PR if we want though. 

(also at some point renaming `OrganisationType` to something like `SourceType` might be helpful.)

src/SFA.DAS.EmployerAccounts.Web/Views/EmployerAccount/Summary.cshtml
Howesy77: Moved into model

src/SFA.DAS.EmployerAccounts.Web.UnitTests/Orchestrators/SearchPensionRegulatorOrchestratorTests/WhenISearchThePensionRegulator.cs
```diff
@@ -58,6 +60,19 @@ public async Task ThenThePensionRegulatorOrganisationListIsReturnedInTheResult()
 
             //Assert
             Assert.IsAssignableFrom<OrchestratorResponse<SearchPensionRegulatorResultsViewModel>>(actual);
-        }       
+        }
+
+        [Test]
+        public async Task ThenEachResultIsCorrectlyMakredAsComingFromPensionsRegulator()
```
MichaelYoung1981: Typo in test name - Makred

src/SFA.DAS.EmployerAccounts.Web/Orchestrators/SearchPensionRegulatorOrchestrator.cs
MichaelYoung1981: I agree it shouldn't be here although disagree that it should come from the service.  I'd say it only needs to be set by whatever method is pushing it back to be saved.

Winsto: That would make sense, although I have a fear this will be used as a 'flag' to switch between options somewhere later; not sure, but would not be surprised.

PR #1696, opened 2019-06-12
AML-3504-common-flag-for-all-AORN-stories

Winsto:
Quickly set one flag as per discussion in chat.


Winsto:
We have decided we will make this change as part of each story. First one in wins :0)

PR #1698, opened 2019-06-13
CON-602 Refactoring

lazcool:
Create extension methods on our types.
Also remove some old files.


src/SFA.DAS.EAS.Portal.Worker/TypesExtensions/AccountExtensions.cs
```diff
@@ -0,0 +1,34 @@
+using System.Linq;
+using SFA.DAS.EAS.Portal.Client.Types;
+
+namespace SFA.DAS.EAS.Portal.Worker.TypesExtensions
+{
+    public static class AccountExtensions
+    {
+        public static (Organisation, EntityCreation) GetOrAddOrganisation(this Account account, long accountLegalEntityId)
```
Stephek: Can we change this to remove the need for the caller to handle the EntityCreation.  So add optional parameters of Action created and Action updated  i.e.

public static Organisation GetOrAddOrganisation(this Account account, long accountLegalEntityId, Action created = null, Action updated = null).

Then the caller can decide if it needs to handle these scenarios and we don't need the EntityCreation enum.

lazcool: Good Idea. Done.

For consistency, I've created actions for the get and add on the helpers, even if they aren't currently used. (`GetOrAddOrganisation` has a consumer that needs the add action, but the other helpers don't currently have consumers for both get and add).

lazcool:
> Could do with some unit tests for the extension methods.

Unit tests added.

PR #1699, opened 2019-06-13
AML-3688 Gateway Inform content change

Howesy77:


src/SFA.DAS.EmployerAccounts.Web/SFA.DAS.EmployerAccounts.Web.csproj
```diff
@@ -701,9 +701,7 @@
     <Content Include="Views\EmployerAccountPaye\Index.cshtml" />
     <Content Include="Views\EmployerAccountPaye\NextSteps.cshtml" />
     <Content Include="Views\EmployerAccountPaye\Remove.cshtml" />
-    <Content Include="Views\Shared\GatewayInform.cshtml" />
```
tom-gough: looks like you have updated gateway inform but removed it from the csproj?

Howesy77: Duplicates

PR #1705, opened 2019-06-14
Con 603 separate event handler concerns

Stephek:


Ashley-Bass:
All looks pretty good. Might be worth going through and renaming a few things as it's a tad confusing having an event handler pass an event through to a handler which then handles that same event.

src/SFA.DAS.EAS.Portal/Application/Services/AccountDocumentServiceWithDuplicateCheck.cs
```diff
@@ -19,6 +19,11 @@ public Task<AccountDocument> Get(long id, CancellationToken cancellationToken =
             return _accountDocumentService.Get(id, cancellationToken);
         }
 
+        public Task<AccountDocument> GetOrCreate(long id, CancellationToken cancellationToken = default)
```
Ashley-Bass: Would it be worth splitting the Get/GetOrCreate and the Save out into separate interfaces to stop duplication of code?

src/SFA.DAS.EAS.Portal/Application/Services/AccountDocumentServiceWithSetProperties.cs
```diff
@@ -19,6 +19,11 @@ public Task<AccountDocument> Get(long id, CancellationToken cancellationToken =
             return _accountDocumentService.Get(id, cancellationToken);
         }
 
+        public Task<AccountDocument> GetOrCreate(long id, CancellationToken cancellationToken = default)
```
Ashley-Bass: Would it be worth splitting the Get/GetOrCreate and the Save out into separate interfaces to stop duplication of code?

src/SFA.DAS.EAS.Portal/Application/Services/AccountDocumentService.cs
```diff
@@ -20,6 +20,17 @@ public Task<AccountDocument> Get(long id, CancellationToken cancellationToken =
                .SingleOrDefaultAsync(a => a.AccountId == id, cancellationToken);            
         }
 
+        public async Task<AccountDocument> GetOrCreate(long id, CancellationToken cancellationToken = default)
```
Ashley-Bass: Would it be worth splitting the Get/GetOrCreate and the Save out into separate interfaces to stop duplication of code?

src/SFA.DAS.EAS.Portal/Application/Services/AccountDocumentServiceWithDuplicateCheck.cs
Stephek: The clients would have to inject both the interfaces though.

Ashley-Bass: But it could mean we could only inject one where necessary if we only needed to get or save to the repo. It's preference either way in my opinion

Stephek: Every client does a GetOrCreate and then a Save.

Ashley-Bass: Worth keeping as is then 👍

lazcool:
We've gone a very roundabout way to effectively rename the old commands from XCommand to XHandler! :smile:

Personally, I'm not convinced of the merits of splitting the event handler classes into two paired event handler classes. I thought your original suggestion of merging the old commands into the eventhandlers was a good one.

src/SFA.DAS.EAS.Portal.Worker/EventHandlers/ProviderRelationships/AddedAccountProviderEventHandler.cs
```diff
@@ -1,74 +1,26 @@
...
 using SFA.DAS.EAS.Portal.Application.Services;
-using SFA.DAS.EAS.Portal.Worker.TypesExtensions;
 using SFA.DAS.ProviderRelationships.Messages.Events;
-using SFA.DAS.Providers.Api.Client;
 
 namespace SFA.DAS.EAS.Portal.Worker.EventHandlers.ProviderRelationships
 {
-    //todo: subscribe to provider updated event (is there 1?) and update our local store provider details
-    
-    /// <remarks>
```
lazcool: Do you believe these notes provide no value, or do you not believe in documentation in general?

I can understand the removal of the section on 'Options for retrieving provider details' somewhat, as it pertains to a decision made, as opposed to providing context to the existing code. However, the 'Notes on handling provider already exists' does provide context to the code in the method (now moved into the Handler), that you can't get from just reading the code.

src/SFA.DAS.EAS.Portal.UnitTests/Worker/EventHandlers/EventHandlerTestsFixture.cs
```diff
@@ -23,119 +16,36 @@ public class EventHandlerTestsFixture<TEvent, TEventHandler>
         public TEvent OriginalMessage { get; set; }
         public IHandleMessages<TEvent> Handler { get; set; }
         public string MessageId { get; set; }
-        public Mock<IMessageContextInitialisation> MessageContextInitialisation { get; set; }
+        public Mock<IMessageContextInitialisation> MockMessageContextInitialisation { get; set; }
```
lazcool: You've added `Mock` to the start of the variable name, but not to the other mocked property. I'm not too bothered about whether to prefix with `Mock` or not, but now we have inconsistent naming.

src/SFA.DAS.EAS.Portal.Worker/EventHandlers/ProviderRelationships/AddedAccountProviderEventHandler.cs
Stephek: If it's a TODO either do it or add a tech debt item.  If you need excessive comments to understand the code then there is a problem with the code.  Notes on general design decisions like the "Notes on handling provider already exists" would be better in a readme file.  Looking at the code the notes offer no value to this unit of code.  The unit tests should provide that value.

Stephek:
> We've gone a very roundabout way to effectively rename the old commands from XCommand to XHandler! 😄
> 
> Personally, I'm not convinced of the merits of splitting the event handler classes into two paired event handler classes. I thought your original suggestion of merging the old commands into the eventhandlers was a good one.

I've implemented what was agreed.

src/SFA.DAS.EAS.Portal.UnitTests/Worker/EventHandlers/EventHandlerTestsFixture.cs
Stephek: Fixed

Stephek:
> All looks pretty good. Might be worth going through and renaming a few things as it's a tad confusing having an event handler pass an event through to a handler which then handles that same event.

One is at the transport layer, the other the domain layer. i.e. in different namespaces.  We have separated the concerns of handling the event in the domain (the use case) from the mechanism that we receive the events.

src/SFA.DAS.EAS.Portal.Worker/EventHandlers/ProviderRelationships/AddedAccountProviderEventHandler.cs
lazcool: Personally, I value lowering the bar for developers new to the code base, over saving space at the top of the file, that can be easily ignored by developers already familiar with the system.

lazcool:
> > We've gone a very roundabout way to effectively rename the old commands from XCommand to XHandler! 😄
> > Personally, I'm not convinced of the merits of splitting the event handler classes into two paired event handler classes. I thought your original suggestion of merging the old commands into the eventhandlers was a good one.
> 
> I've implemented what was agreed. "Split NSB Event Handlers app and Domain Event Handles"

Perhaps next time we can use refactor -> rename. Should save time and doesn't remove the git history for the code.

src/SFA.DAS.EAS.Portal.UnitTests/Worker/EventHandlers/EventHandlerTestsFixture.cs
cofaulco: a note on this, I always suffix with Mock so that there is a distinction but intellisense doesn't pick up everything starting 'Mock' when you know which you want

Stephek: Changed to be consistent.

src/SFA.DAS.EAS.Portal.UnitTests/Portal/Application/EventHandlerTestsFixture.cs
```diff
@@ -0,0 +1,31 @@
+﻿using AutoFixture;
+using SFA.DAS.EAS.Portal.Application.EventHandlers;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace SFA.DAS.EAS.Portal.UnitTests.Portal.Application
+{
+    public class EventHandlerTestsFixture<TEvent, TEventHandler>
```
lazcool: All the common code in the event handler test fixture base class has been duplicated by copy and pasting into each event handler's test fixture. We should move the common code back into the common test fixture. Duplicating the code makes maintenance more time consuming.

PR #1710, opened 2019-06-18
First try, just change config to not errror.

Winsto:


Winsto:
This is not the change we are looking for.

PR #1717, opened 2019-06-20
AML-3614 Removed SFA.DAS.Extensions

Howesy77:


src/SFA.DAS.EmployerAccounts.Web/Extensions/StringExtensions.cs
```diff
@@ -12,30 +32,91 @@ public static string FormatPayeForUrl(this string scheme)
...
             return scheme.Replace("_", "/");
+        }
+
+        public static bool IsEquivalent(this string s1, string s2)
+        {
+            return s1.IsEquivalent(s2, StringEquivalenceOptions.Default);
         }
 
-        public static decimal? AsNullableDecimal(this string input)
+        public static bool IsEquivalent(this string s1, string s2, StringEquivalenceOptions options)
```
Winsto: I realise this is just a "lift and shift" from the `SFA.DAS.Extensions` code, but something doesn't seem quite right about adding a string extension for this.

There is only one client site calling this, specifically for the case of comparing organisation names in `OrganisationOrchestrator.CheckForUpdate`.

Two suggestions for re-factors spring to mind: -

One: 

If this is a quirk of organisation names that needs to be considered in multiple places than introduce a value (DDD sense) type to represent that and then either override equality in the type or implement an `IEqualityComparer<OrganisationName>`

Two:

As this is only used in one place (at the moment) make it a private 'utility' method of `OrganisationOrchestrator`.

The private method can be done a little more elegantly than this imperative implementation.

Option one: -

```
currentValue
.Replace(" ", String.Empty)
.Equals(
updatedValue
.Replace(" ", String.Empty),
StringComparison.Ordinal)
```

This is, imho, more elegant and also allows us to specify `Ordinal` or `OrdinalIgnoreCase` options. It isn't the most efficient (if we really care in this use case) as we are generating extra strings with the `Replace` (current imperative implementation is comparing chars). If this matters,

Option two: -
```
currentValue
.Where(c => ! char.IsWhiteSpace(c))
.SequenceEqual(
  updatedValue
  .Where(c => ! char.IsWhiteSpace(c))
```

we do not get to specify case sensitivity of comparison.

Howesy77: Moved UnitTests into Web project. Kept current implementation as can't be sure of reasoning behind current implementation.

PR #1720, opened 2019-06-25
Aml 3724

Howesy77:


src/SFA.DAS.EAS.Employer_Account.Database/StoredProcedures/AddPayeToAccount.sql
```diff
@@ -4,19 +4,20 @@
 	@accessToken VARCHAR(30),
 	@refreshToken VARCHAR(30),
 	@addedDate DATETIME,
-	@employerRefName VARCHAR(500) NULL
+	@employerRefName VARCHAR(500) NULL,
+	@aorn VARCHAR(50) NULL
```
Winsto: On the TPR api side AORN is varchar(25), but then from [what I can see](https://www.gov.uk/pay-paye-tax/bank-details) AORN is 13 chars, so don't know if there is a good reason for varchar(25).

Howesy77: Changed to VARCHAR(25)

Winsto:
Updated target to [feature branch](https://github.com/SkillsFundingAgency/das-employerapprenticeshipsservice/tree/AML-3707-AML-3706-Account-Levy-Status)

PR #1722, opened 2019-06-25
Aml 3726 publish event

Winsto:
Publish event when levy transaction amount greater than zero.


src/SFA.DAS.EmployerFinance.UnitTests/Events/ProcessDeclarationTests/WhenIProcessDeclarations.cs
```diff
@@ -49,5 +56,39 @@ public async Task ThenTheLoggerIsCalledWithInfoLevel()
             //Assert
             _logger.Verify(x=>x.Info("Process Declarations Called"));
         }
+
+        [TestCase(100023,0.0,false)]
+        [TestCase(900034,10.0,true)]
+        public async Task ThenLevyAddedToAccountEventIsPublishedForGreaterThanZeroLevyAmount(
```
MichaelYoung1981: I would split this into 2 tests (positive and negative) as it's confusing otherwise.

Winsto: Split into three (levyAmount greater then, less than and equal to zero) now.

Oh and this is just straight equality `levyAmount == decimal.zero` do we need to consider tolerance equality?

I haven't seen examples of tolerance equality in the code, but thought it worth asking.

src/SFA.DAS.EmployerFinance.UnitTests/Events/ProcessDeclarationTests/WhenIProcessDeclarations.cs
```diff
@@ -49,5 +58,98 @@ public async Task ThenTheLoggerIsCalledWithInfoLevel()
...
             _logger.Verify(x=>x.Info("Process Declarations Called"));
         }
+
+        [Test]
+        public async Task ThenLevyAddedToAccountEventIsPublishedForGreaterThanZeroLevyAmount()
+        {
+            long accountId = 900034;
+            decimal levyAmount = 10.0m;
+
+            _dasLevyRepository
```
MichaelYoung1981: Can the duplicated code from the 3 tests be refactored out so we don't have 3 copies?

Winsto: Yes it can; I sort of thought about that as I was copying and pasting, but went with the lazy option :)

PR #1724, opened 2019-06-26
Aml 3699 exclude aorn accounts from levy processing

Winsto:
This includes the change from https://github.com/SkillsFundingAgency/das-employerapprenticeshipsservice/tree/AML-3724
AML-3724 can go into master ahead of this change, but we need the new field to filter out AORN accounts in this branch.

Note: this a re-work of the first go at this; in this version I have left the original sproc and repository method `IEmployerSchemesRepository.GetSchemesByEmployerId` and added new code to filter out AORN added PAYE schemes.


Winsto:
> Given that sub-tasks aren't testable in isolation and we can't merge to master before a branch has been tested, shouldn't we be making changes onto a branch for the user story (if we want to use sub branches that get merged into that branch after code review, fair enough)?  Otherwise you're going to have to pull all the branches together before it can be tested anyway so we're just making things more complicated for ourselves by splitting them out in the first place.

For the Levy status work the idea was to have task branches back to a "feature" branch.

For this work, filtering out paye schemes added via AORN route from levy processing; I thought it could be tested in isolation (isolation from the other Levy status work that is), so made this PR back to master.

However I am also happy to merge this back to the feature branch for Levy status work if that is better for testing.

Winsto:
Updated target to [Levy status feature branch](https://github.com/SkillsFundingAgency/das-employerapprenticeshipsservice/tree/AML-3707-AML-3706-Account-Levy-Status)

PR #1726, opened 2019-06-26
Aml 3728 consume event

Winsto:
PR back to feature branch for consuming event.


src/SFA.DAS.EAS.Employer_Account.Database/Tables/Account.sql
```diff
@@ -5,7 +5,8 @@
     [Name] NVARCHAR(100) NOT NULL ,
 	[CreatedDate] DATETIME NOT NULL,
 	[ModifiedDate] DATETIME NULL, 
-    [PublicHashedId] NVARCHAR(100) NULL
+    [PublicHashedId] NVARCHAR(100) NULL, 
+    [LevyType] TINYINT NOT NULL DEFAULT 0
```
MichaelYoung1981: LevyType needs to be renamed (although I think what you have is a better name but ApprenticeshipEmployerType was agreed)

PR #1735, opened 2019-07-05
CON-662 Vacancy API

Ashley-Bass:
https://skillsfundingagency.atlassian.net/browse/CON-662

Add a call to the vacancy API to the portal client.


src/SFA.DAS.EAS.Portal.Client/Configuration/RecruitApiClientConfiguration.cs
```diff
@@ -0,0 +1,14 @@
+﻿using SFA.DAS.Http;
+
+namespace SFA.DAS.EAS.Portal.Client.Configuration
+{
+    public class RecruitApiClientConfiguration : IAzureADClientConfiguration
+    {
+        public string ApiBaseUrl { get; set; }
+        public string TimeoutTimeSpan { get; set; }
```
cofaulco: Is it worth this being a timeout in seconds which we convert to a timespan in the code?

lazcool: The advantage of using a string that is fed to `TimeSpan.Parse`, is that it allows us to use any value that `TimeSpan` supports. E.g. we could set the timeout to 2.5 seconds.

src/SFA.DAS.EAS.Portal.Client/IPortalClient.cs
```diff
@@ -6,6 +6,7 @@ namespace SFA.DAS.EAS.Portal.Client
 {
     public interface IPortalClient
     {
-        Task<Account> GetAccount(long accountId, CancellationToken cancellationToken = default);
+        Task<Account> GetAccount(long accountId, string publicHashedAccountId,
+            bool hasPayeScheme, CancellationToken cancellationToken = default);
```
cofaulco: I think we talked about this point earlier on. Creating a class of filters with a HasPayeScheme boolean property. This will allow it to be easily extended if more flags are added in the future.

Also could we not has the id in the client rather than passing it in alongside the accountId ?

src/SFA.DAS.EAS.Portal.Client.TestHarness/DependencyResolution/IoC.cs
```diff
@@ -5,9 +5,14 @@ namespace SFA.DAS.EAS.Portal.Client.TestHarness.DependencyResolution
 {
     public static class IoC
     {
-        public static void Initialize(Registry registry)
+        public static IContainer Initialize()
```
Stephek: Good tidy up.

src/SFA.DAS.EAS.Portal.Client.TestHarness/DependencyResolution/LoggerRegistry.cs
```diff
@@ -0,0 +1,13 @@
+using SFA.DAS.NLog.Logger;
+using StructureMap;
+
+namespace SFA.DAS.EAS.Portal.Client.TestHarness.DependencyResolution
+{
+    public class LoggerRegistry : Registry
+    {
+        public LoggerRegistry()
+        {
+            For<ILog>().Use(c => new NLogLogger(c.ParentType, null, null)).AlwaysUnique();
```
Stephek: Might be better to use a NullLogger or FakeLogger for tests.

src/SFA.DAS.EAS.Portal.Client/Application/Queries/IGetAccountQuery.cs
```diff
@@ -0,0 +1,11 @@
+using System.Threading;
+using System.Threading.Tasks;
+using SFA.DAS.EAS.Portal.Client.Types;
+
+namespace SFA.DAS.EAS.Portal.Client.Application.Queries
+{
+    internal interface IGetAccountQuery
```
Stephek: Can this be named IAccountQuery ? The Get is a method of the interface so doesn't need to be in the interface name.

src/SFA.DAS.EAS.Portal.Client/IPortalClient.cs
lazcool: I've changed the signature to accept a flags enum, so that we if we need to pass more flags, we don't have to add further parameters. It also has the bonus of making the consumer code more readable, because it will pass `AccountState.HasPayeScheme` rather than `true`.

I've also moved decoding of the hashed `AccountId` into the client, so now the method only has to accept a hashed `AccountId`, rather than the hashed and vanilla versions.

src/SFA.DAS.EAS.Portal.Client/Application/Queries/IGetAccountQuery.cs
lazcool: Currently, we only have one query, but I envisage that each query will be contained within its own object, rather than having multiple query methods within a type ([CQRS, rather than CQS](http://cqrs.nu/Faq)). In which case, I think it's more appropriate to name the class and interface after the query, rather than the entity/DTO, as we could have multiple queries for a single entity/DTO, in which case there would be naming collisions,

src/SFA.DAS.EAS.Portal.Client/PortalClient.cs
```diff
@@ -1,23 +1,58 @@
...
+
+            // at a later date, we might want to create an empty account doc and add the vacancy details to it, but for now, let's keep it simple
+            if (!hasPayeScheme || account == null)
+                return account;
+            
+            var vacancies = await vacanciesTask;
+            if (vacancies == null)
+                return account;
+            
+            var vacanciesCount = vacancies.Count();
```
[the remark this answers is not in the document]
lazcool: Agreed, it makes the code more concise and just as readable. Change made.

src/SFA.DAS.EAS.Portal.Client/PortalClient.cs
```diff
@@ -1,23 +1,58 @@
...
+            switch (vacanciesCount)
+            {
+                case 0:
+                    account.VacancyCardinality = Cardinality.None;
+                    break;
+                case 1:
+                    account.VacancyCardinality = Cardinality.One;
+                    account.SingleVacancy = vacancies.Single();
```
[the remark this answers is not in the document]
lazcool: I think this is a worthwhile optimisation (and improves the interface). Previously, this PR requested up to a 1000 vacancy records, mapped them all, then returned them all to the consumer, when all the consumer needs to know is if there is 0, 1 or >1 vacancies, and the details of a single vacancy, if there is only one.

What it now does is return an enum stating f there are 0, 1 or >1 vacancies and the details of a maximum of 1 vacancy.

This means that instead of the vacancy API having to read up to 1000 full records from its local store, return those records over the network, then the client storing them in memory, mapping them to the required subset of properties, then passing them to the consumer (and then all that memory having to be garbage collected), the vacancy API only has to read and send a maximum of two records, the client only maps a maximum of two records (could be 1!) and then returns a maximum of 1 record subset.

This should have a considerable impact on the time taken to render the homepage, and the reduction in memory usage and churn should also enable the website and vacancy API to perform its other actions quicker.

It also makes the interface more intuitive, as it wasn't obvious before that it could be returning only a subset of the entire vacancy set, whereas now it's more obvious what is being returned.

If we wanted to to return the full set of vacancies, we'd have to write extra code to page through the vacancies returned from the API, adding unnecessary complexity.

It's also easy to change the interface if requirements change. The vacancy records are retrieved from the API each time and added to the account fetched from the document collection. We don't store the vacancies in the account document in the CosmosDB document collection, so we don't have to worry about backwards compatibility while reading the account, or updating existing account documents if the schema changes.

If requirements change, we can change the interface to match the new requirements. We might, for example, want to display a maximum of 5 records in a tabbed interface, but I doubt if we'd ever want to display hundreds of vacancies on the homepage, at least not without paging, so I doubt if we'll ever need to return the full set of vacancies.

src/SFA.DAS.EAS.Portal.Client/Types/Account.cs
```diff
@@ -18,6 +18,10 @@ public Account()
         public ICollection<Organisation> Organisations { get; set; }
         [JsonProperty("providers")]
         public ICollection<Provider> Providers { get; set; }
+        [JsonIgnore]
+        public Cardinality? VacancyCardinality { get; set; }
```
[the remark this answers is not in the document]
lazcool: I refer the honourable developer to the answer I gave a moment ago 😄.

src/SFA.DAS.EAS.Portal/DependencyResolution/CommitmentsService.cs
```diff
@@ -27,6 +27,9 @@ public class CommitmentsApiClientConfiguration : ICommitmentsApiClientConfigurat
             public string BaseUrl { get; set; }
             public string ClientToken { get; set; }
         }
+        
+        //todo: as we're not in the client, we could/should use RestHttpClient here instead
```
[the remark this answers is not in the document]
lazcool: The `todo` has been removed and a tech debt task added to the backlog.

src/SFA.DAS.EAS.Portal.Client.TestHarness/DependencyResolution/LoggerRegistry.cs
lazcool: I find it useful to view the log from the test harness, to review the appropriateness of our logging.

src/SFA.DAS.EAS.Portal.Client/Configuration/RecruitApiClientConfiguration.cs
lazcool: I've also introduced a rexeg in the config's json schema, to check any supplied value for correctness, to address your spoken concern of incorrectly formatted values not being picked up in the config system.

src/SFA.DAS.EAS.Portal.Client/Application/Queries/IGetAccountQuery.cs
Stephek: We don't have commands since we don't have a write model so CQRS doesn't apply.  My comment was about the naming of the interface.  Do we even need this interface?  The concrete is just a helper method for simplifying the query against the repository.  In that case its a query service so a better name would IAccountQueries or IAccountQueryService with an initial method GetActiveAccount() [or GetAccount()  with it implied that the account is active]  since that is whet this helper object is doing.  Then we may add methods like GetDeletedAccount() or FindByName()  methods etc...

src/SFA.DAS.EAS.Portal.Client/Services/DasRecruit/DasRecruitService.cs
```diff
@@ -0,0 +1,72 @@
...
+    {
+        private readonly ILog _log;
+        private readonly HttpClient _httpClient;
+        private readonly Type _vacancyStatusType;
+
+        public DasRecruitService(IRecruitApiHttpClientFactory recruitApiHttpClientFactory, ILog log)
+        {
+            _log = log;
+            //todo: think through lifetimes
+            _httpClient = recruitApiHttpClientFactory.CreateHttpClient();
```
Stephek: Should httpClient be injected in as a singleton as per here?

 https://docs.microsoft.com/en-gb/dotnet/api/system.net.http.httpclient?view=netframework-4.7.1

src/SFA.DAS.EAS.Portal.Client/PortalClient.cs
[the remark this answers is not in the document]
Stephek: This is implementing a UI requirement in the back end rather than supporting the requirements of the UI.

I agree with Corey that the model returned shouldn't have this UI rule in it.  The original model had a collection of Vacancies in it which meets the requirements.  The cardinality is implicit by the number of records in that collection so we don't need this VacancyCardinality property on the account model.  As far as the back end is concerned (being a service) the UI requires all the records back for that account (since it may be populating a drop down list so the required vacancy can be selected).  If there is only one vacancy record then only one will be returned so it performs for that scenario.  This is not an optimisation but a corruption of the model, optimisations should only be done when there is a requirement for them based on empirical measurements.  A better approach to optimise would be to include in the Filter object how many records to return and a value specifying if there are more records to return.  i.e. Filter.MaxVacancy = -1 .  So then the client can decide if an optimisation is required since it can set the value to -1 (All), 1, 3 or 4 or to what ever the UI requirement is and therefore is no longer constrained by the back end implementation.  In this case a count of the total number of records would be useful to return.  By taking this approach if the UI requirements change then the back end can support those requirements without being changed.  Likewise a full or summary vacancy resultset can be returned by setting another filter property.

src/SFA.DAS.EAS.Portal.Client/PortalClient.cs
```diff
@@ -1,23 +1,63 @@
...
-        public Task<Account> GetAccount(long accountId, CancellationToken cancellationToken = default)
+        public async Task<Account> GetAccount(string hashedAccountId,
+            AccountState accountState, CancellationToken cancellationToken = default)
         {
-            return _getAccountQuery.Get(accountId, cancellationToken);
+            var hasPayeScheme = (accountState & AccountState.HasPayeScheme) == AccountState.HasPayeScheme;
+
+            // we potentially map 1 more vacancy than necessary, but it keeps the code clean
+            var vacanciesTask = hasPayeScheme ? 
+                _dasRecruitService.GetVacancies(hashedAccountId, 2, cancellationToken) : null;
```
Stephek: Can the decorator pattern be used here to separate the concerns of Getting the vacancy data and merging with the account from the concerns of getting the account model.  Then the existing code does not need to be changed and the new code is independent and can be tested separately and removed/modified if the source of vacancy data changes.

src/SFA.DAS.EAS.Portal.Client/IPortalClient.cs
```diff
@@ -6,6 +6,7 @@ namespace SFA.DAS.EAS.Portal.Client
 {
     public interface IPortalClient
     {
-        Task<Account> GetAccount(long accountId, CancellationToken cancellationToken = default);
+        Task<Account> GetAccount(string hashedAccountId, AccountState accountState,
```
Stephek: This interface is a query interface.  Basically we are asking for an account filtered by some data.  AccountState with 
 a HasPayeScheme property is a bit meaningless in this instance.  What the client really wants is to get vacancy data included in the resultset.  Think of it as doing a search and limiting or expanding the results of the search.  The service just needs to be told what to return.  A filter with an "IncludeVacancyData = true" property or similar is more intuitive to a client than "HasPayeScheme = true".  In the context of a Get request HasPayeScheme isn't a clear request without knowing about a business rule that may change.  As a client I would expect HasPayeScheme to return the same data whether true or false with the difference being that when true only an account with a PayeScheme would be returned/filtered.

src/SFA.DAS.EAS.Portal.Client/Application/Queries/IGetAccountQuery.cs
lazcool: Yes, the client really has no need for queries or commands. It's likely that it'll only have to fetch an account, so we might as well keep it simple and just have a method on the repository. That allows us to remove some infrastructure that wasn't adding any value.

src/SFA.DAS.EAS.Portal.Client/PortalClient.cs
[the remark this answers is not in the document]
cofaulco: I like the idea of extending the filter and adding the max vacancies property. I think this retains the optimisation that Phil has introduced without changing the account schema based on current requirements. This would mean there would be little change required should the consumer of the client or the source of the Vacancy data change.

src/SFA.DAS.EAS.Portal.Client/PortalClient.cs
cofaulco: Agree although not a showstopper in my opinion. Could add a small tech debt task to cleanup

src/SFA.DAS.EAS.Portal.Client/IPortalClient.cs
cofaulco: I agree with IncludeVacancyData > HasPayeScheme. I think this is a fair point

src/SFA.DAS.EAS.Portal.Client/Application/Queries/IGetAccountQuery.cs
cofaulco: I don't mind either way on the naming of this one.

src/SFA.DAS.EAS.Portal.Client/IPortalClient.cs
cofaulco: 👍  on the hashing, I think there are comments further down to potentially argue that a simple class with different flags on might be more appropriate, if we have IncludeVacancyData = true AND MaxVacancies = 2

src/SFA.DAS.EAS.Portal.Client/PortalClient.cs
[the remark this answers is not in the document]
lazcool: I have addressed your concerns in the order that you raised them...

Yes, if the client returned the full set of Vacancies, then the `VacancyCardinality` property would not be needed, but by not returning the full set, it definitely is.

The front end does not require all the records back, as proven by the requirements being met by the client only returning a maximum of 1 record. The UI is *not* populating a select list, and I think it extremely unlikely that we'll ever be asked to populate a select list with up to a thousand options. If we *were* ever tasked to implement that, then we are free to change the code to match any new requirements.

Yes, if there is only one vacancy, then performance will be fine. It was not that scenario that I was worried would slow down the rendering of the homepage.

I agree in principle with not optimising prematurely, and optimizing based on empirical measurements, but common sense has to prevail and we should never fetch a thousand records when we only need one. Just because you shouldn't optimise prematurely, doesn't mean that you should make the system do extra unnecessary work, and I don't need to measure results to know that fetching two records is going to be considerably quicker than fetching a 1000!

With regards to your suggestion of passing a filter to the client, that would involve implementing paging in the service that talks to the vacancy API, and implementing paging in the client itself, all for some potential future requirement that may never come. We should keep the code as simple as possible to implement the current requirements, and if new requirements come along, we can adapt the system to best fit the *actual* new requirements, rather than some hypothetical requirements. A new requirement may be to replace the content of the panel currently radiating vacancy information, with information from another system. In that case, implementing two sets of paging code would have been a waste of time and money.

I have the same issue with your suggestion of passing a flag to the client to get it to return either a full or summary vacancy record. There is no requirement to return a full vacancy record, so there is no point in spending time implementing such functionality. That road leads to an unnecessarily complex system because future requirements have been guessed at. The unnecessary complexity also makes it harder to refactor the system for actual new requirements.

The considerations would be different if we were implementing a general library or framework, where you have to guess at consumer requirements, but we're effectively implementing a service for a single consumer, where both are released together.

src/SFA.DAS.EAS.Portal.Client/Types/Account.cs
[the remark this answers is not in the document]
lazcool: I'm resolving this conversation, as the discussion is taking place in the thread above.

src/SFA.DAS.EAS.Portal.Client/PortalClient.cs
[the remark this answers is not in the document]
cofaulco: We could apply a page size of whatever the filter max vacancies property is set to which would avoid the need to implement paging logic. This way we would get the benefits mentioned in my previous comment. Setting the value to 2 in the caller would give us all the information we need currently from vacancies. If a time came when a consumer asks for a thousand vacancies we would optimise and add paging at that point.

This I think would be a good compromise with fairly minimal changes.

src/SFA.DAS.EAS.Portal.Client/Services/DasRecruit/DasRecruitService.cs
lazcool: That's what the `todo` was there for, to make sure that the HttpClient is a singleton 😄.

I've made the `DasRecruitService` a singleton, which effectively makes `HttpClient` a singleton.

src/SFA.DAS.EAS.Portal.Client/PortalClient.cs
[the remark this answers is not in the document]
lazcool: So basically, you're suggesting to move the logic about how many vacancies to fetch, and the cardinality calculation, out of the client and up into the website. (That would effectively mean moving roughly half of the client's `GetAccount` method into the website.)

So the debate has basically boiled down to where the logic associated with the portal should live. 

I personally think it's cleaner to encapsulate such knowledge in the portal's code, rather than having it in the controller/website, (we already have domain knowledge in the portal), but if you want the logic moved into the website, I can move it.

(Having the logic in the website also has the disadvantage of returning more data than is required, but only up to twice the number of vacancies, rather than a 1000 😄.)

src/SFA.DAS.EAS.Portal.Client/IPortalClient.cs
lazcool: The reason I went with `hasPayeScheme` over e.g. `includeVacancyData`, is that we can potentially reuse `hasPayeScheme` to apply other optimisations, based on whether the account has a paye scheme or not. If we go with instructional flags, (such as `includeVacancyData`), then we could end up sending multiple flags, that are all either set to true or false as a group (effectively rendering all but one of the flags redundant).

It also puts the logic in the client (which I think of as a service for the website), rather than in the website/controller.

The flags enum is named `AccountState` to indicate that that it represents the state of the account. If it was a filter as you expected it to be, I would have called it something like `FilterBy`. I've added XML documentation to the client's public interface, to describe the purpose of the parameters, which intellisense should pop up.

src/SFA.DAS.EAS.Portal.Client/PortalClient.cs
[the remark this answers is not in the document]
cofaulco: I think it makes more sense for the website (consumer of the client) to decide how to use the data that it is presented. The logic itself would be encapulated within the viewmodel then as it derives the cardinality and therefore which view to display from the returned vacancies.

src/SFA.DAS.EAS.Portal.Client/PortalClient.cs
lazcool: I personally think that while doing what you suggest would separate concerns, it would also add unnecessary complexity. Remember that `PortalClient` is injected by the client, not on the portal side, so the decoration would have to be done within the website's DI container. If the website's DI library is changed, which could be out of our control, the new library might not support decoration.

We could add another layer within the client's code to inject a new account service and decorate that, but then we'd have to modify the existing code anyway.

src/SFA.DAS.EAS.Portal.Client/IPortalClient.cs
lazcool: The correct location of the client related logic is being debated in other threads, and the outcome of those threads will inform the interface.

src/SFA.DAS.EAS.Portal.Client/IPortalClient.cs
cofaulco: I think the point still stands here that the consumer of our client doesn't necessarily know what they are getting or omitting by setting the account state whereas a filter of IncludeVacancyData would be more explicit

src/SFA.DAS.EAS.Portal.Client/PortalClient.cs
cofaulco: I thought that we provide a PortalClientRegistry so would the decorations not be setup in here. If we replaced StructureMap this would cease to work as is regardless wouldn't it ?

src/SFA.DAS.EAS.Portal.Client/Application/Queries/IGetAccountQuery.cs
lazcool: A `Get` method has been added to the accounts repository, and the query removed.

src/SFA.DAS.EAS.Portal.Client/PortalClient.cs
lazcool: We do, and yes we'd have to change the portal code to support a different DI library.

src/SFA.DAS.EAS.Portal.Client/IPortalClient.cs
lazcool: That's the beauty of encapsulation, the client doesn't need to know the details of why it gets everything it needs, it just does. I can change the flag's name to `IncludeVacancies`, or whatever you wish, if that's your call.

cofaulco: I just think that coming at it as a new dev you wouldn't have a clue why you were getting no vacancy data and that you need to know the business rule of "HasPayeScheme must be true in order to get Vacancy Data"

src/SFA.DAS.EAS.Portal.Client/PortalClient.cs
cofaulco: all I am getting at is that there are further implications if we were to change the DI library and we possibly shouldn't let that stop us from using decorators

src/SFA.DAS.EAS.Portal.Client/PortalClient.cs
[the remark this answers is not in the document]
Stephek: The portal is a read only model - the client is just allowing a client to query that data.  If the portal received the vacancy data via events would you design the interface in this way?  Would you still pass a HasPayeScheme in?  You are leaking implementation details rather than encapsulating them.

src/SFA.DAS.EAS.Portal.Client/IPortalClient.cs
lazcool: As long as the the client passes the correct state of the account, they will get vacancy data (if there is any). The client _does not_ need to know the business rule, if the rule is contained within the client. They _will_ need to know the business rule, if it's contained in the client!

lazcool: Actually, if the client passes the maximum number of vacancies it wants, there's no needs to pass a flag, it can just say it want 0 vacancies.

cofaulco: If the business rule changes though then both the caller and the portal client would need to be changed to set/read new setting rather than simply changing where IncludeVacancyData is set on the consumer end?

src/SFA.DAS.EAS.Portal.Client/Services/DasRecruit/DasRecruitService.cs
cofaulco: 👍

src/SFA.DAS.EAS.Portal.Client/PortalClient.cs
```diff
@@ -41,8 +41,15 @@ public PortalClient(IContainer container, IEncodingService encodingService)
...
             // at a later date, we might want to create an empty account doc and add the vacancy details to it, but for now, let's keep it simple
             if (vacanciesTask != null && account != null)
-                account.Vacancies = (await vacanciesTask).ToList();
-            
+            {
+                var vacancies = await vacanciesTask;
+                if (vacancies != null)
+                {
+                    account.VacanciesRetrieved = true;
+                    account.Vacancies = (await vacanciesTask).ToList();
```
cofaulco: We could use the vacancies variable which was awaited from the task above

src/SFA.DAS.EAS.Portal.Client/IPortalClient.cs
```diff
@@ -10,10 +10,10 @@ public interface IPortalClient
         /// Get the account details required to render the homepage.
         /// </summary>
         /// <param name="hashedAccountId">The (non-public) hashed account id.</param>
-        /// <param name="accountState">The current state of the account. Set all flags that apply.</param>
+        /// <param name="maxNumberOfVacancies">The maximum number of vacancies to return.</param>
         /// <param name="cancellationToken">An optional cancellation token.</param>
         /// <returns></returns>
-        Task<Account> GetAccount(string hashedAccountId, AccountState accountState,
+        Task<Account> GetAccount(string hashedAccountId, int maxNumberOfVacancies,
```
cofaulco: Again I think this should be a filter class of some sort with a int MaxNumberOfVacancies property to make this extensible without changing the signature of the method all the time.

src/SFA.DAS.EAS.Portal.Client/IPortalClient.cs
lazcool: Moving the business logic up the stack, does mean that any change to the logic has the possibility of impacting more of the stack.

However, in this case, I think that any code changes required from the business logic changing, should be able to be contained to the website. If it requires vacancy records, it can say how many it wants, and if it doesn't need any, it can pass 0, with no changes to the portal client required.

src/SFA.DAS.EAS.Portal.Client/IPortalClient.cs
lazcool: I've now introduced a (business) parameter object for the `GetAccount` method, as agreed offline.

src/SFA.DAS.EAS.Portal.Client/PortalClient.cs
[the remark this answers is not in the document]
lazcool: Deciding how many vacancies to fetch and counting the vacancies, has been moved out of the portal client (and the related website PR's will be refactored to include that code), and a parameter object has been introduced to pass the business parameters to the `GetAccount` method (as agree offline).

src/SFA.DAS.EAS.Portal.Client/PortalClient.cs
lazcool: A debate for another day 😄.

src/SFA.DAS.EAS.Portal.Client/IPortalClient.cs
lazcool: The interface has now been agreed.

src/SFA.DAS.EAS.Portal.Client/PortalClient.cs
lazcool: Done (the code was wip 😉).

src/SFA.DAS.EAS.Portal.Client/IPortalClient.cs
lazcool: A parameter object has now been introduced and the interface is agreed - unblocking the front end stories.

PR #1744, opened 2019-07-10
UNTAGGLE: fix for missing config.

Winsto:


src/SFA.DAS.EmployerAccounts/DependencyResolution/TasksRegistry.cs
```diff
@@ -11,7 +11,8 @@ public class TasksRegistry : Registry
     {
         public TasksRegistry()
         {
-            For<ITaskApiConfiguration>().Use(c => c.GetInstance<IAutoConfigurationService>().Get<ITaskApiConfiguration>(ConfigurationKeys.TasksApi)).Singleton();
+            For<ITaskApiConfiguration>().Use<TaskApiConfiguration>();
```
MichaelYoung1981: This line shouldn't be necessary, the only change should be removing the I from the Get<>.  I've made the change directly on the Untaggle branch to get a build through.

Winsto:
Closing as this work has been done directly on the branch now.

PR #1746, opened 2019-07-12
CON-667 Vacancy live status web front end

lazcool:


cofaulco:
Merged as part of feature branch

PR #1749, opened 2019-07-18
CON-735 Remove saved favourites link

lazcool:
https://skillsfundingagency.atlassian.net/browse/CON-735

Note: we don't remove everything that's only there for the saved favourites link (e.g. related model code & image), as it's expected to come back soon.


lazcool:
Something's gone crazy with this PR. Closing and will create another one.

PR #1753, opened 2019-07-20
AML-3603 Block Aorn Route

Howesy77:


src/SFA.DAS.EAS.Employer_Account.Database/StoredProcedures/UpdateUserAornAttempts.sql
```diff
@@ -0,0 +1,23 @@
+﻿CREATE PROCEDURE [employer_account].[UpdateUserAornAttempts]
+(
+	@UserRef UNIQUEIDENTIFIER,
+	@Suceeded BIT
```
MichaelYoung1981: Spelling mistake

src/SFA.DAS.EmployerAccounts.UnitTests/Queries/GetUserAornLockTests/WhenIGetUserAornLockStatus.cs
```diff
@@ -0,0 +1,48 @@
...
+        {
+            _userAornPayeLockService = new Mock<IUserAornPayeLockService>();
+            _userAornPayeLockService.Setup(x => x.UserAornPayeStatus(It.IsAny<Guid>())).ReturnsAsync(new UserAornPayeStatus());
+
+            Query = new GetUserAornLockRequest(Guid.NewGuid().ToString());
+            RequestHandler = new GetUserAornLockQueryHandler(_userAornPayeLockService.Object);
+        }
+
+        [Test]
+        public async Task ThenIfTheServiceIsCalled()
```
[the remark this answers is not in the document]
MichaelYoung1981: Dodgy test name in general. :)  It's very implementation specific, it'd be better if it described the behaviour (ThenLockStatusIsReturned?) which would mean both tests could be squashed (and still have a single assertion.

src/SFA.DAS.EmployerAccounts.UnitTests/Queries/UpdateUserAornLockTests/WhenIUpdateUserAornLockStatus.cs
```diff
@@ -0,0 +1,36 @@
...
+        {             
+            _userAornPayeLockService = new Mock<IUserAornPayeLockService>();
+            _userAornPayeLockService.Setup(x => x.UpdateUserAornPayeAttempt(It.IsAny<Guid>(), It.IsAny<bool>())).ReturnsAsync(true);
+
+            Query = new UpdateUserAornLockRequest(Guid.NewGuid().ToString(), true);
+            RequestHandler = new UpdateUserAornLockQueryHandler(_userAornPayeLockService.Object);
+        }
+
+        [Test]
+        public async Task ThenIfTheMessageIsValidTheRepositoryIsCalled()
```
MichaelYoung1981: Can you make the test name less implementation specific (ThePayeAttemptIsUpdated?)

src/SFA.DAS.EmployerAccounts.UnitTests/Services/UserAornPayeLockServiceTests.cs
```diff
@@ -0,0 +1,144 @@
...
+                UserAornPayeLock = new UserAornPayeLockConfiguration
+                {
+                    NumberOfPermittedAttempts = testData.NumberOfPermittedAttempts,
+                    PermittedAttemptsTimeSpanMinutes = testData.PermittedAttemptsTimeSpanMinutes,
+                    LockoutTimeSpanMinutes = testData.LockoutTimeSpanMinutes
+                }
+            };
+
+            var service = new UserAornPayeLockService(userRepo.Object, config, logger);
+            var result = service.UserAornPayeStatus(It.IsAny<Guid>()).Result;
```
MichaelYoung1981: Can we set a GUID rather than using IsAny?

src/SFA.DAS.EmployerAccounts.UnitTests/Services/UserAornPayeLockServiceTests.cs
```diff
@@ -0,0 +1,144 @@
...
+                Expected = new UserAornPayeStatus { RemainingLock = 29, IsLocked = true, RemainingAttempts = 0, AllowedAttempts = 3 }
+            }
+        };
+     
+        [Test]
+        public void ItShouldReturnTheCorrectAornLockStatus([ValueSource("TestData")] TestData testData)
+        {
+            var logger = Mock.Of<ILog>();
+            var userRepo = new Mock<IUserRepository>();
+            userRepo.Setup(x => x.GetAornPayeQueryAttempts(It.IsAny<Guid>())).ReturnsAsync(testData.Attempts.ToList());
```
MichaelYoung1981: Can we set a GUID rather than using IsAny?

src/SFA.DAS.EmployerAccounts.Web/Controllers/EmployerAccountPayeController.cs
```diff
@@ -21,18 +20,18 @@ public class EmployerAccountPayeController : BaseController
         private const int AddPayeUsingAorn = 2;
 
         private readonly EmployerAccountPayeOrchestrator _employerAccountPayeOrchestrator;
+        private readonly UserAornLockOrchestrator _userAornLockOrchestrator;
```
MichaelYoung1981: Looking at the number of methods this is used in I'm not convinced it should be added as a dependency as most methods don't need it.  If you look at how many places have had to be changed to simply pass a new mock object it highlights the issue.  Could we do it using an attribute?

src/SFA.DAS.EmployerAccounts.Web/Controllers/SearchPensionRegulatorController.cs
```diff
@@ -126,19 +142,33 @@ public async Task<ActionResult> SearchPensionRegulatorByAorn(SearchPensionRegula
         [NonAction]
```
MichaelYoung1981: I know you haven't added them but what's the point of these attributes on a private method?

src/SFA.DAS.EmployerAccounts.Web/Orchestrators/UserAornLockOrchestrator.cs
```diff
@@ -0,0 +1,40 @@
...
+using AutoMapper;
+using MediatR;
+using SFA.DAS.EmployerAccounts.Models.UserProfile;
+using SFA.DAS.EmployerAccounts.Queries.GetUserAornLock;
+using SFA.DAS.EmployerAccounts.Queries.UpdateUserAornLock;
+using SFA.DAS.EmployerAccounts.Web.ViewModels;
+
+namespace SFA.DAS.EmployerAccounts.Web.Orchestrators
+{
+    public class UserAornLockOrchestrator
```
MichaelYoung1981: I don't think this is an orchestrator.  The (truly awful) pattern has a 1-2-1 relationship between controllers and orchestrators.

src/SFA.DAS.EmployerAccounts/Data/UserRepository.cs
```diff
@@ -127,5 +128,33 @@ public async Task<Users> GetAllUsers()
...
+            var parameters = new DynamicParameters();
+            parameters.Add("@UserRef", userRef, DbType.Guid);
+
+            var query = await _db.Value.Database.Connection.QueryAsync<DateTime>(
+                sql: "[employer_account].[GetUserAornAttempts]",
+                param: parameters,
+                transaction: _db.Value.Database.CurrentTransaction.UnderlyingTransaction,
+                commandType: CommandType.StoredProcedure);
+
+            return query.ToList();
```
MichaelYoung1981: Won't the query just return a single result?

src/SFA.DAS.EmployerAccounts/Services/UserAornPayeLockService.cs
```diff
@@ -0,0 +1,87 @@
...
+            catch (Exception e)
+            {
+                _logger.Error(e, "An error occurred when calling the Aorn Lock Service.");
+                return false;
+            }
+            
+            return true;
+        }
+
+        public async Task<UserAornPayeStatus> UserAornPayeStatus(Guid userRef)
```
MichaelYoung1981: This method could do with a quick refactor to extract a few bits out to make it more readable.

src/SFA.DAS.EmployerAccounts.Web/Controllers/EmployerAccountPayeController.cs
```diff
@@ -21,18 +20,18 @@ public class EmployerAccountPayeController : BaseController
...
-            IAuthenticationService owinWrapper, 
+            IAuthenticationService owinWrapper,
             EmployerAccountPayeOrchestrator employerAccountPayeOrchestrator,
-            IAuthorizationService authorization, 
+            UserAornLockOrchestrator userAornLockOrchestrator,
+            IAuthorizationService authorization,
             IMultiVariantTestingService multiVariantTestingService,
             ICookieStorageService<FlashMessageViewModel> flashMessage) : base(owinWrapper, multiVariantTestingService, flashMessage)
         {
-            if (owinWrapper == null)
```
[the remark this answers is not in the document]
Howesy77: Yeah tidied up code. These tests aren't required as StructureMap would fail before we got to the point.

src/SFA.DAS.EmployerAccounts/Data/UserRepository.cs
Howesy77: No it returns a list of the attempts.

MichaelYoung1981: Makes sense given the name of the method.

PR #1756, opened 2019-07-21
AML-3711 Content warning

Howesy77:


Howesy77:
It won't show

Winsto:
Raised question about js disabled behaviour on ticket and as per conversation in standup, will wait to see if this is the solution we want before approving.

Winsto:
Closing as after user research and discussion we need a different solution to this problem.

PR #1761, opened 2019-07-23
CON-767 Show relevant finance links to levy and non levy

lazcool:
https://skillsfundingagency.atlassian.net/browse/CON-767

Branched off CON-482-interim-homepage.
Includes AML-3707-AML-3706-Account-Levy-Status, which includes change to API to return levy/non-levy flag.


src/SFA.DAS.EmployerAccounts/DependencyResolution/AccountApiClientRegistry.cs
```diff
@@ -0,0 +1,16 @@
+﻿using SFA.DAS.EAS.Account.Api.Client;
+using StructureMap;
+
+namespace SFA.DAS.EmployerAccounts.DependencyResolution
+{
+    public class AccountApiClientRegistry : Registry
```
lazcool: I've added this registry here, rather than in the client project, so that we don't introduce a dependency on StructureMap in the client.

src/SFA.DAS.EmployerAccounts/DependencyResolution/AccountApiClientRegistry.cs
```diff
@@ -0,0 +1,16 @@
...
+    {
+        // we add this here, rather than in the client project, so we don't introduce a dependency on structuremap in the client
+        public AccountApiClientRegistry()
+        {
+            // the client's not very good, but the version of sfa.das.http currently in EAS isn't either, so we go with the client
+            For<IAccountApiClient>().Use<AccountApiClient>();
```
lazcool: The client's not very good, but the version of `SFA.DAS.Http` currently in EAS isn't either, so we go with the client.

src/SFA.DAS.EmployerAccounts.Web/Orchestrators/EmployerTeamOrchestrator.cs
```diff
@@ -32,12 +34,14 @@ public class EmployerTeamOrchestrator : UserVerificationOrchestratorBase
     {
         private readonly IMediator _mediator;
         private readonly ICurrentDateTime _currentDateTime;
+        private readonly IAccountApiClient _accountApiClient;
 
-        public EmployerTeamOrchestrator(IMediator mediator, ICurrentDateTime currentDateTime)
+        public EmployerTeamOrchestrator(IMediator mediator, ICurrentDateTime currentDateTime, IAccountApiClient accountApiClient)
```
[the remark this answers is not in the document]
lazcool: The `AccountApiClient` is used to query. We're not issuing a command to it, so a command handler is not appropriate.

Looking at examples of where an API is used to query data, employer accounts currently uses...

`IReferenceDataApiClient`, which is used by the `ReferenceDataService`
`ITokenServiceApiClient`, which is used by the `HmrcService`
`IEmployerCommitmentApi`, which is used by `PaymentService` and `CommitmentService`

So if anything, introducing a service would be more appropriate. However, introducing a service to only wrap a call to `GetAccount`, without any manipulation of the parameters or response, wouldn't add any value, apart from consistent layering. I personally prefer to be pragmatic and not introduce layers that don't add any value. As the code evolves however, it may make sense to introduce a service to wrap the account client.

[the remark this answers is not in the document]
Stephek: Ok so query handlers rather than command handlers (an unnecessary pedantic point). 

In the GetAccount method of the orchestrator the mediator is used to get data with 

GetEmployerAccountHashedQuery
GetTeamMemberQuery
GetAccountStatsQuery
GetUserAccountRoleQuery
GetAccountEmployerAgreementsRequest

So therefore for consistency the same approach should be used to get the accountApi data.  So it's not more appropriate to introduce a different way of returning the data.  This is not about personal opinion it is about following the existing patterns as already implemented.  "Consistently bad is better than inconsistently good." Consistency makes the code easier to understand and work with.

src/SFA.DAS.EmployerAccounts.Web/Orchestrators/EmployerTeamOrchestrator.cs
```diff
@@ -198,9 +204,14 @@ public async Task<OrchestratorResponse<AccountDashboardViewModel>> GetAccount(st
                     RequiresAgreementSigning = pendingAgreements.Count(),
                     AgreementsToSign = pendingAgreements.Count() > 0,
                     SignedAgreementCount= agreementsResponse.EmployerAgreements.Count(x => x.HasSignedAgreement),
-                    PendingAgreements = pendingAgreements
+                    PendingAgreements = pendingAgreements,
+                    ApprenticeshipEmployerType = (ApprenticeshipEmployerType)Enum.Parse(typeof(ApprenticeshipEmployerType), apiGetAccountTask.Result.ApprenticeshipEmployerType, true)
                 };
 
+                //note: ApprenticeshipEmployerType is already returned by GetEmployerAccountHashedQuery, but we need to transition to calling the api instead.
```
Stephek: Do we need this comment?  It seems obvious you may want to parse to an enum.

lazcool: Yes, I think it's an important comment. Without it, a developer might notice that `ApprenticeshipEmployerType` is returned by both the query and the api call, and remove the call to the api call as it's effectively redundant (no other property is used from the api call).

However, we've been asked to pick up the flag from the api, as we want to transition to retrieving data through the api, rather than going direct to the database, so it wouldn't be appropriate to remove the api call.

src/SFA.DAS.EmployerAccounts.Web/Orchestrators/EmployerTeamOrchestrator.cs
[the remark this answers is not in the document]
lazcool: I think _it is_ important to be precise when coding and discussing code. If i'd changed the code as you suggested, by introducing a command handler to query an API, that would have been the wrong thing to do (which we both agree on), and it would have taken time to implement it, and then extra time to correct it.

You are talking about consistency with what's already in the orchestrator's method, but consistency with how existing API's are called is key. I'll give an example to illustrate my point. If we needed to introduce a command call in the method, we wouldn't put it in a query because the method currently calls queries, instead we'd make it consistent with other commands in the system.

The same applies with introduding an API call. Throughout the system, API clients are wrapped by services, and those services are injected into orchestrators.

So the argument for consistency (which is an eminently sensible argument), should be to integrate the API in a consistent manner, rather than to use a query because the `GetAccount` method already calls queries.

However, on balance, I think we should only introduce a service to wrap the API, when it would, in itself, add some value.

PR #1762, opened 2019-07-23
Aml 3762 eoi api

DannyGB:
Includes work to add EoI information to the existing accounts API


src/SFA.DAS.EAS.Account.Api.Types/AgreementViewModel.cs
```diff
@@ -9,5 +9,6 @@ public class AgreementViewModel
         public string SignedByName { get; set; }
         public EmployerAgreementStatus Status { get; set; }
         public int TemplateVersionNumber { get; set; }
+        public string AgreementType { get; set; }
```
cofaulco: Jumping on this one as we are going to be consuming this flag. Is it worth this property being an enum so that we aren't string matching all over the place?

DannyGB: A fair point however the guidance I have from the architects is that it "should be set to an appropriate (parsable) value; e.g. NonLevy.EOI.1" so I used string. Although the documentation doesn't state why it should be parsable I'm reluctant to change it without understanding the reasoning being it.

Do you have in mind a use-case that will require us to perform logic on this field internally? If not consumers of the API can avoid string matching by converting the value to an Enum when they retrieve the data.

Howesy77: I think it'll be used as part of the UI EOI implementation. The Levy flag is defined in SFA.DAS.Common.Domain.Types. Should the Agreement Type Enum be there too?

DannyGB: Please see [this page](https://skillsfundingagency.atlassian.net/wiki/x/B4AtVQ) for the reasoning behind this implementation

MichaelYoung1981:
Ignore me closing the pr, I pressed the wrong button. 😄

src/SFA.DAS.EAS.Account.Api.Types/AgreementViewModel.cs
MichaelYoung1981: Could we split out type and version? That’d give everyone what they want as we wouldn’t need any code updates when a new version is introduced and we’d need no magic string parsing. 
Having this as a string throughout is inconsistent with how we manage pretty much everything else. 
How does versioning on a levy agreement work? Is there already a column for that or does it use a substring from the template?

DannyGB: @MichaelYoung1981 There's already a version number on the DB for the agreement type template which Levy uses however there is a unique index on only the version column. I don't want to add another column to that index as that has the potential to break existing functionality.

Also the EOI work will be largely throw away so ultimately it'll be a trade off between the amount of work we put in to make the solution "right" vs the short-term nature of it and also the need to unblock the other teams that have dependencies on the code base

Stephek: @DannyGB you don't need to add another column.  Just update the AgreementMappings 

.ForMember(v => v.AgreementType, o => o.MapFrom(a => a.Template != null ? a.Template.AgreementType : null))

s that they parse into an AgreementInfo object (or similar)

AgreementInfo
{
   AgreementType  Type  {get; set;}  ( Enum Levy/NonLevy}
   AgreementPhase Phase {get; set;}  (null for existing}
}

AgreementPhase
{
   Phase Phase {get; set;} (Enum EOI)
   int Version {get; set;}  (1)
}

Then you are encapsulating the data storage of the data and not polluting the domain with how the data is stored (a physical concern) and are providing the logical view of the domain to consumers of the service and making the code base extensible in the future.  Otherwise all clients will need to know the database format and do the parsing.

PR #1763, opened 2019-07-23
Con 666 664 663 merged with 667

Stephek:


src/SFA.DAS.EmployerAccounts.Web/Views/EmployerTeam/V2/VacancyStatus.cshtml
```diff
@@ -1,24 +1,58 @@
...
+    }
 
     <dt class="das-definition-list__title">Status</dt>
-    <dd class="das-definition-list__definition">@vacancy.Status</dd>
+    <dd class="das-definition-list__definition">@status</dd>
 </dl>
 <ul class="govuk-list">
-    <li><a href="@vacancy.ManageVacancyUrl" class="govuk-link govuk-link--no-visited-state">Manage vacancy</a></li>
-    <li><a href="@Url.EmployerRecruitAction("create-options/")" class="govuk-link govuk-link--no-visited-state">Create a vacancy</a></li>
+    <li><a href="@vacancy.ManageVacancyUrl" class="govuk-link govuk-link--no-visited-state">@actionLinkText</a> </li>
```
cofaulco: I think this needs to point to a specific action or at least supply the id of the vacancy which we are attempting to manage ?

src/SFA.DAS.EmployerAccounts.Web/Views/EmployerTeam/V2/VacancyStatus.cshtml
```diff
@@ -1,24 +1,58 @@
...
+    };
+
+    if (Model.HasSingleDraftVacancy)
+    {
+        vacancy = Model.AccountViewModel.Vacancies.First(v => v.Status == VacancyStatus.Draft);
+        actionLinkText = "Edit and submit vacancy";
+        status = "Draft";
+    };
+
+    if (Model.HasSingleRejectedVacancy)
```
cofaulco: this looks like it includes work for CON-665 Rejected Vacancy

src/SFA.DAS.EmployerAccounts.Web/Views/EmployerTeam/V2/VacancyStatus.cshtml
```diff
@@ -1,24 +1,58 @@
...
 @{
-    var vacancy = Model.AccountViewModel.SingleVacancy;
-}
+    Vacancy vacancy = null;
+
+    var actionLinkText = "";
+    int? numberOfApplications = null;
+    string status = "";
+
+    if (Model.HasSingleClosedVacancy)
```
cofaulco: could the Row2COlumn2 action call a helper to create a "VacancyViewModel" which we could pass into this partial and therefore test this logic ?

src/SFA.DAS.EmployerAccounts.Web/Views/EmployerTeam/V2/VacancyStatus.cshtml
Stephek: Yes that's correct

src/SFA.DAS.EmployerAccounts.Web/Views/EmployerTeam/V2/VacancyStatus.cshtml
Stephek: Not sure about that.  Was told in a previous review comment that this is all that is required.

src/SFA.DAS.EmployerAccounts.Web/Views/EmployerTeam/V2/VacancyStatus.cshtml
```diff
@@ -1,24 +1,58 @@
...
+    };
+
+    if (Model.HasSingleDraftVacancy)
+    {
+        vacancy = Model.AccountViewModel.Vacancies.First(v => v.Status == VacancyStatus.Draft);
+        actionLinkText = "Edit and submit vacancy";
+        status = "Draft";
+    };
+
+    if (Model.HasSingleRejectedVacancy)
```
lazcool: The way the `HasSingleXxxVacancy` properties on the model are calculated, it looks like you could have e.g. 4 vacancies, 1 each of the 4 statuses checked, so all the properties are set to true, then here, the last status checked would be true (`HasSingleRejectedVacancy`), so the 4 returned vacancies would be interpreted as a single rejected vacancy.

The changes to this view also breaks the existing handling of the draft status vacancy.

This panel is only shown when there is a single vacancy, so all it has to do is check `vacancy.Status` (and the `HasSingleXxxVacancy` properties can be removed.

Stephek: The panel is only called from the controller if there is a single vacancy.  The rule in the view are just UI rules.

lazcool: Yes, I realised that after writing the comment, so what's here should work.

We could tidy it up by taking the first vacancy, and using it's status, but that's a tidy up, as opposed to an issue.

src/SFA.DAS.EmployerAccounts.Web/Views/EmployerTeam/V2/VacancyStatus.cshtml
```diff
@@ -1,24 +1,58 @@
...
+    {
+        vacancy = Model.AccountViewModel.Vacancies.First(v => v.Status == VacancyStatus.Referred);
+        actionLinkText = "Edit and re-submit vacancy";
+        status = "Rejected";
+    };
+}
 <h3 class="das-panel__heading">Your vacancy</h3>
-<p>@vacancy.TrainingTitle <span class="das-!-display--block das-text--muted">VAC@(vacancy.Reference)</span></p>
+<p>@vacancy.TrainingTitle <span class="das-!-display--block das-text--muted">@("VAC" + vacancy.Reference)</span></p>
```
lazcool: We don't actually need the string concatenation

src/SFA.DAS.EmployerAccounts.Web/Views/EmployerTeam/V2/VacancyStatus.cshtml
```diff
@@ -1,24 +1,58 @@
...
+    @{ if (numberOfApplications.HasValue)
+        {
+            <dt class="das-definition-list__title">Applications</dt>
+            <dd class="das-definition-list__definition">@numberOfApplications.Value</dd>
+        }
+    }
 
     <dt class="das-definition-list__title">Status</dt>
-    <dd class="das-definition-list__definition">@vacancy.Status</dd>
+    <dd class="das-definition-list__definition">@status</dd>
```
lazcool: We could just use `@vacancy.Status`.

src/SFA.DAS.EmployerAccounts.Web/Views/EmployerTeam/V2/VacancyStatus.cshtml
Stephek: I've refactored to tidy up.

PR #1764, opened 2019-07-23
Con 770 skip sign agreement

cofaulco:
Check agreement signed status as well as feature toggle and PAYE/Org added to determine routing to new/old homepage


src/SFA.DAS.EmployerAccounts.Web/Controllers/EmployerTeamController.cs
```diff
@@ -482,14 +482,21 @@ private async Task<OrchestratorResponse<AccountDashboardViewModel>> GetAccountIn
...
-        private bool HasOrganisation(AccountDashboardViewModel data)
+        private bool HasOrganisation(AccountDashboardViewModel accountViewModel)
         {
-            return data.OrgainsationCount > 0;
+            return accountViewModel.OrganisationCount > 0;
+        }
+
+        private bool DisplayNewHomepage(AccountDashboardViewModel accountViewModel)
+        {
+            return FeatureToggles.Features.HomePage.Enabled
```
lazcool: Can you confirm that we should display the new homepage even when the the feature isn't enabled?

cofaulco: yes the logic is, new homepage should be displayed at all times when toggled on i.e in DEMO.

When it is toggled off we want to direct the users to new homepage to prompt them to add their PAYE (when skipped) and to prompt them to sign their agreement (new)

lazcool: 👍

src/SFA.DAS.EmployerAccounts.Web.UnitTests/Controllers/EmployerTeamControllerTests/WhenAgreementsToSign.cs
```diff
@@ -64,5 +64,24 @@ public void ThenTheProviderPermissionsDeniedViewIsReturned()
...
+
+            model.AccountViewModel = new Model.Account();
+            model.AccountViewModel.Providers.Add(new Model.Provider());
+
+            //Act
+            var result = _controller.Row2Panel2(model) as PartialViewResult;
+
+            //Assert
+            Assert.IsNotNull(result);
+            Assert.AreEqual("PreAgreementRecruitment", (result.Model as dynamic).ViewName);
```
lazcool: Should cast model to `PanelViewModel<AccountDashboardViewModel>`, rather than `dynamic`.

lazcool:
This PR was accidentally merged into master. There is now a new [PR](https://github.com/SkillsFundingAgency/das-employerapprenticeshipsservice/pull/1820) with these changes.

PR #1766, opened 2019-07-25
CON-769 Finance navigation on interim homepage

lazcool:


src/SFA.DAS.EmployerFinance/DependencyResolution/ConfigurationRegistry.cs
```diff
@@ -11,6 +12,7 @@ public ConfigurationRegistry()
             For<EmployerAccountsConfiguration>().Use(c => c.GetInstance<IAutoConfigurationService>().Get<EmployerAccountsConfiguration>(ConfigurationKeys.EmployerAccounts)).Singleton();
             For<EmployerFinanceConfiguration>().Use(c => c.GetInstance<IAutoConfigurationService>().Get<EmployerFinanceConfiguration>(ConfigurationKeys.EmployerFinance)).Singleton();
             For<ForecastingApiClientConfiguration>().Use(c => c.GetInstance<IAutoConfigurationService>().Get<ForecastingApiClientConfiguration>(ConfigurationKeys.ForecastingApiClient)).Singleton();
+            For<IAccountApiConfiguration>().Use(c => c.GetInstance<EmployerAccountsConfiguration>().AccountApi).Singleton();
```
lazcool: Both accounts web and finance web, both read in the config for accounts and finance. Most config seems to be in accounts config however, so we pick up the account API config from there, which already contains it (also the config is account related).

src/SFA.DAS.EmployerFinance.Web/Extensions/UrlHelperExtensions.cs
```diff
@@ -22,6 +22,14 @@ public static string EmployerCommitmentsAction(this UrlHelper helper, string pat
             return AccountAction(helper, baseUrl, path);
         }
 
+        public static string ReservationsAction(this UrlHelper helper, string path)
```
lazcool: Didn't end up using this, but it should be there.

PR #1767, opened 2019-07-25
AML-3762 Added EOI whitelist flag to DB for legal entity creation.

Howesy77:
Passes the whitelist flag to the DB to select the correct Agreement Type.
Contains branch AML 3762-EOI-API for DB changes.


src/SFA.DAS.EmployerApprenticeshipsService.Domain/Data/Repositories/IAccountRepository.cs
```diff
@@ -11,7 +11,7 @@ namespace SFA.DAS.EAS.Domain.Data.Repositories
     public interface IAccountRepository
     {
         Task AddPayeToAccount(Paye payeScheme);
-        Task<CreateAccountResult>  CreateAccount(long userId, string employerNumber, string employerName, string employerRegisteredAddress, DateTime? employerDateOfIncorporation, string employerRef, string accessToken, string refreshToken, string companyStatus, string employerRefName, short source, short? publicSectorDataSource, string sector, string aorn);
+        Task<CreateAccountResult> CreateAccount(long userId, string employerNumber, string employerName, string employerRegisteredAddress, DateTime? employerDateOfIncorporation, string employerRef, string accessToken, string refreshToken, string companyStatus, string employerRefName, short source, short? publicSectorDataSource, string sector, string aorn, bool Eoi);
```
Winsto: That is a long list of parameters :) starting to look like maybe a request type could be worth introducing.

Not required change just commenting as I had to scroooooooolll across to see the diff at the end of the line :)

MichaelYoung1981: Probably better if the repository takes the domain object it’s trying to save. 👍

Howesy77: Some calls have been implemented as a request type.

The command calls a number of SP's

Winsto:
@MichaelYoung1981 I thought we had completed an untangle task to replace local `SFA.DAS.Authorization` with the shared version; it looks like not. I take it we are continuing with the locacl/legacy `SFA.DAS.Authorication` even when adding new work (like this feature whitelist) and then undoing everything as part of an untangle task?

src/SFA.DAS.EAS.Employer_Account.Database/StoredProcedures/CreateEmployerAgreement.sql
```diff
@@ -10,7 +11,7 @@ BEGIN
 	BEGIN
 		SELECT TOP 1 @templateId = Id
 		FROM [employer_account].[EmployerAgreementTemplate]
-		WHERE AgreementType = 'Levy'
+		WHERE AgreementType LIKE CASE @eoi WHEN 0 THEN 'Levy' ELSE 'NonLevy.EOI%' END
```
Winsto: Is this related to the 'magic string' discussion that was had around the api work?

Also, but this is wider than just this code review, are we happy that we could set `AgreementType` to 'NonLevy' here, but may well set the account Levy status to 'Levy' via the completely independent mechanism that sets that other flag?

src/SFA.DAS.EmployerAccounts.UnitTests/Commands/CreateAccountCommandTests/WhenICallCreateAccount.cs
```diff
@@ -58,7 +58,7 @@ public void Arrange()
...
-            _accountRepository.Setup(x => x.CreateAccount(It.IsAny<long>(), It.IsAny<string>(), It.IsAny<string>(), It.IsAny<string>(), It.IsAny<DateTime?>(), It.IsAny<string>(), It.IsAny<string>(), It.IsAny<string>(), It.IsAny<string>(), It.IsAny<string>(), It.IsAny<short>(), It.IsAny<short?>(), It.IsAny<string>(), It.IsAny<string>())).ReturnsAsync(new CreateAccountResult { AccountId = ExpectedAccountId, LegalEntityId = ExpectedLegalEntityId, EmployerAgreementId = ExpectedEmployerAgreementId, AccountLegalEntityId = ExpectedAccountLegalEntityId });
+            _accountRepository.Setup(x => x.CreateAccount(It.IsAny<long>(), It.IsAny<string>(), It.IsAny<string>(), It.IsAny<string>(), It.IsAny<DateTime?>(), It.IsAny<string>(), It.IsAny<string>(), It.IsAny<string>(), It.IsAny<string>(), It.IsAny<string>(), It.IsAny<short>(), It.IsAny<short?>(), It.IsAny<string>(), It.IsAny<string>(), It.IsAny<bool>())).ReturnsAsync(new CreateAccountResult { AccountId = ExpectedAccountId, LegalEntityId = ExpectedLegalEntityId, EmployerAgreementId = ExpectedEmployerAgreementId, AccountLegalEntityId = ExpectedAccountLegalEntityId });
```
Winsto: We could save ourselves a lot of pain in these test setups if we refatored that long list of parameters to a request type. Or as @MichaelYoung1981 said used the domain object.

src/SFA.DAS.EAS.Employer_Account.Database/StoredProcedures/CreateEmployerAgreement.sql
Howesy77: This scenario will be caught in the web app.

src/SFA.DAS.EmployerApprenticeshipsService.Domain/Data/Repositories/IAccountRepository.cs
Winsto: If I comment out the lines setting the `Eoi` property in `EmployerAccountController` and `EmployerAccountOrchestrator` then all the tests under `SFA.DAS.EmployerAccounts/Tests` still pass? There does not appear to be any coverage of the logic for setting this flag. 

On that point I don't see why we are setting it in the controller; from what I can see it is a value that we need to persist based purely on the whitelist status of the user with no rules for deciding this in controller or orchestrator.

I can see that we do not, necessarily, want to inject `IAuth..Service` into command handlers to have it set there; preferring the handlers to process the request/command that was passed to them. But if feels like something that is outside of any steps we have. We could maybe use `Mediatr` pre-request pipeline to augment the request based on whitelist status.

Another approach might be to inject a command factory into, say, `EmployerAccountOrchestrator`. So, for example, in `EmployerAccountOrchestrator.CreateNewAccount` instead of...

```
                var result = await Mediator.SendAsync(new CreateAccountCommand
                {
                    ExternalUserId = model.UserId,
                    OrganisationType = model.OrganisationType,
                    OrganisationName = model.OrganisationName,
.
.
.
                    Eoi = model.Eoi
```

we have something like : -

```
            try
            {
                var command = commandFactory.GetCreateAccountCommand();

                command.ExternalUserId = .....

```

or

```
            try
            {
                var command = commandFactory.GetCreateAccountCommandForCreateAccountModel(model);

```

and push the authorization checking into that factory.

We could also just have a model -> command mapper.

Wherever we put that logic we should cover it with tests and having the parameter set at the  _top of the chain_ in the controller is just going to me a lot more tests modifying than we need.

                var result = await Mediator.SendAsync(new CreateAccountCommand
                {
                    ExternalUserId = model.UserId,

Winsto:
Looking at the ticket for this there are scenarios talking about changing the flow of pages based on the EOI whitelist status. This change seems to be only about persisting the EOI status, I don't see any switching logic based on it, is this PR for a sub-task of that story?

src/SFA.DAS.EAS.Employer_Account.Database/StoredProcedures/CreateEmployerAgreement.sql
MichaelYoung1981: We’re using the template (which has a version column on it) but we’re adding the version onto the end of the type and then doing like checks in stored procs? Is this some kind of elaborate wind up?

src/SFA.DAS.EmployerAccounts.Web/Controllers/EmployerAccountController.cs
```diff
@@ -49,6 +52,7 @@ public class EmployerAccountController : BaseController
             _returnUrlCookieStorageService = returnUrlCookieStorageService;
             _accountCookieStorage = accountCookieStorage;
             _hashedAccountIdCookieName = typeof(HashedAccountIdModel).FullName;
+            _authorizationService = authorizationService;
```
Winsto: We are only injecting the dependency on `IAuthorizatoin` servcie to set a property on a model which we than pass to the _next layer down_, this seems a bit redudant. See other comment on lack of test coverage on property setting in this controller.

src/SFA.DAS.EAS.Employer_Account.Database/StoredProcedures/CreateEmployerAgreement.sql
Howesy77: This is in the design as signed off by the TA's.

Howesy77: SFA.DAS.Authorization hasn't been migrated. CI are also using the same for their change.

Winsto: @MichaelYoung1981 Was this magic string thing part of the debate on the [api PR](https://github.com/SkillsFundingAgency/das-employerapprenticeshipsservice/pull/1762)?

In which case is there much we can do about it here?

MichaelYoung1981: I’ll speak to Roy on Monday but Dave’s requirement was simply that he didn’t want consumers of the api to require changes when a new version is introduced.
I’m aware you’re change is based on a previous stories implementation but that implementation shouldn’t have gone to master.

Winsto:
@Howesy77 as per discussion thinking about it command handler and quite probably the repository is where we shut put the logic for setting the EOI flag. Nobody else cares about this. In fact it is only really used to set the template stuff in the database.

From a bit of investigation the authorization service should be fine getting the user from the context when injected into the repository.

src/SFA.DAS.EAS.Employer_Account.Database/StoredProcedures/CreateEmployerAgreement.sql
MichaelYoung1981: @winsto yes, we can amend the original implementation if we feel it’s necessary to preserve the quality of the code base.

Winsto: @MichaelYoung1981 We'll hold off on this PR then until that discussion has been had.

Howesy77:
@MichaelYoung1981 yeah I agree. Neil and I deliberated for a while about the best place for it :). We weren't sure if the command handler should add logic to the calling command.

src/SFA.DAS.EmployerAccounts.UnitTests/Commands/CreateAccountCommandTests/WhenICallCreateAccount.cs
Winsto: This can all change as we can move the logic can go directly in the repository. No need to be passing this `EOI` property through all these layers.

Winsto:
> @MichaelYoung1981 yeah I agree. Neil and I deliberated for a while about the best place for it :). We weren't sure if the command handler should add logic to the calling command.

Except it won't actually be modifying the command that is passed in, as there is no need for that `Eoi` property to be on the command. If we are putting the logic into the command handler, rather than the repository ( and I can see the argument for that) then it will be a parameter passed to the repository, but the value of that parameter comes from the authorization service being injected into the command handler. So the create account command gets reverted to what it was before, just the handler now passes an extra parameter to the repository based on logic in the command handler.

Winsto:
Just noting points from discussion on Friday: -

- the lack of logic changing pages is due to the fact that there won't be any; the templates drive that. This work is just to set the template correctly according to the EOI status of the user creating the account.

- There is no need for an `Eoi` property in any models in the controller or orchestrator. The template needs to be set correctly depending on the EOI status of the user. This status is obtained from the authorization service. So we need to inject the authorization service as close to the point of persistence as reasonable. This could be command handler or repository. There could be a reasonable objection to a _business_ rule such as this in the repository (as per @MichaelYoung1981 's comment), so the command handler might be most appropriate. all other references to Eoi can be removed.
- there is an objection to the _stringly typed_ stuff with the templates. This is driven by the api implementatoin. We will wait, before approving this PR, to have a discussion around that api implementation.

src/SFA.DAS.EmployerAccounts.UnitTests/Commands/CreateAccountCommandTests/WhenICallCreateAccount.cs
Howesy77: I have changed that.. A parameter object is used now.

src/SFA.DAS.EAS.Account.Api.UnitTests/Orchestrators/AccountsOrchestratorTests/WhenIGetANonLevyAccount.cs
```diff
@@ -81,23 +83,23 @@ public async Task ThenResponseShouldHaveAccountAgreementTypeSetToNonLevy()
...
+            Assert.AreEqual(agreementType.ToString(), result.Data.AccountAgreementType.ToString());
         }
 
         [Test]
         public async Task ThenResponseShouldHaveAccountAgreementTypeSetToInconsistent()
         {
-            //Arrange
-            string agreementType = "Inconsistent";
+            //Arrange
+            AccountAgreementType agreementType = AccountAgreementType.Inconsistent;
```
dashton82: not use var?

src/SFA.DAS.EAS.Account.Api.UnitTests/Orchestrators/AccountsOrchestratorTests/WhenIGetANonLevyAccount.cs
```diff
@@ -57,14 +59,14 @@ public void Arrange()
         public async Task ThenResponseShouldHaveAccountAgreementTypeSetToNonLevy()
         {
             //Arrange
-            string agreementType = "NonLevy.EOI";
+            AgreementType agreementType = AgreementType.NoneLevyExpressionOfInterest;
```
dashton82: Use var?

src/SFA.DAS.EAS.Api/Orchestrators/AccountsOrchestrator.cs
```diff
@@ -262,13 +262,17 @@ private async Task<TransferAllowance> GetTransferAllowanceForAccount(long accoun
...
         {
             var agreementTypeGroup = accountResult.Account.AccountAgreementTypes?
-                .GroupBy(x => x)
-            ;
+                .GroupBy(x => x);
 
-            return agreementTypeGroup?.Count() > 1 ? "Inconsistent" : agreementTypeGroup?.FirstOrDefault()?.Key;
+            if (agreementTypeGroup == null || !agreementTypeGroup.Any())
+            {
+                return AccountAgreementType.Unknown;
```
dashton82: Is there a test to cover this scenario?

src/SFA.DAS.EmployerAccounts.UnitTests/Commands/CreateAccountCommandTests/WhenICallCreateAccount.cs
```diff
@@ -202,8 +207,7 @@ public async Task WillCallRepositoryToCreateNewAccount()
 
             await _handler.Handle(cmd);
 
-            _accountRepository.Verify(x => x.CreateAccount(_user.Id, cmd.OrganisationReferenceNumber, cmd.OrganisationName, cmd.OrganisationAddress, cmd.OrganisationDateOfInception, cmd.PayeReference, cmd.AccessToken, cmd.RefreshToken, cmd.OrganisationStatus, cmd.EmployerRefName, (short)cmd.OrganisationType, cmd.PublicSectorDataSource, cmd.Sector, cmd.Aorn));
-            
+            _accountRepository.Verify(x => x.CreateAccount(It.IsAny<CreateAccountParams>()));
```
dashton82: Not sure if this is actually verifying what the test has described

src/SFA.DAS.EmployerAccounts.UnitTests/Commands/CreateAccountCommandTests/WhenICallCreateAccount.cs
```diff
@@ -249,7 +253,7 @@ public async Task ThenAnOrganisationCodeIsGeneratedIfOneIsNotSupplied()
 
             //Assert
 
-            _accountRepository.Verify(x => x.CreateAccount(It.IsAny<long>(), It.Is<string>(cd => !string.IsNullOrEmpty(cd)), It.IsAny<string>(), It.IsAny<string>(), It.IsAny<DateTime?>(), It.IsAny<string>(), It.IsAny<string>(), It.IsAny<string>(), It.IsAny<string>(), It.IsAny<string>(), It.IsAny<short>(), It.IsAny<short?>(), It.IsAny<string>(), It.IsAny<string>()), Times.Once);
+            _accountRepository.Verify(x => x.CreateAccount(It.IsAny<CreateAccountParams>()), Times.Once);
```
dashton82: Not sure if this is verifying the test sufficiently

src/SFA.DAS.EAS.Api/Orchestrators/AccountsOrchestrator.cs
MichaelYoung1981: WhenIGetAnAccountWithNoAgreements.cs

PR #1769, opened 2019-07-25
Con 768 EOI remove links

Stephek:


src/SFA.DAS.EmployerAccounts.Web/Controllers/EmployerTeamController.cs
```diff
@@ -50,7 +50,7 @@ public class EmployerTeamController : BaseController
         public async Task<ActionResult> Index(string hashedAccountId, string reservationId)
         {
             var response = await GetAccountInformation(hashedAccountId);
-            if (FeatureToggles.Features.HomePage.Enabled || !HasPayeScheme(response.Data) && !HasOrganisation(response.Data))
+            /*if (FeatureToggles.Features.HomePage.Enabled || !HasPayeScheme(response.Data) && !HasOrganisation(response.Data))
```
cofaulco: don't think this was intentionally committed

src/SFA.DAS.EmployerAccounts.Web/Extensions/AccountDashboardViewModelExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+                return false;
+            }
+
+            return true;
+        }
+
+        public static bool ShowYourFundingReservationsLink(this AccountDashboardViewModel model)
+        {
+            if(
+                model.ApprenticeshipEmployerType == ApprenticeshipEmployerType.NonLevy ||
```
cofaulco: I think this should be AND logic. EmployerType.NonLevy && AgreementType.EOI

src/SFA.DAS.EmployerAccounts.Web/Controllers/EmployerTeamController.cs
lazcool: Code uncommented.

src/SFA.DAS.EmployerAccounts.Web/Extensions/AccountDashboardViewModelExtensions.cs
lazcool: Perhaps the intent was (in effect) `(NonLevy || NonLevy) && EOI`?

Stephek: There are NonLevy/Levy states on the Employer and the Agreement.  The two are mutually exclusive according to the requirements.  So it is show link if Employer is non levy OR show if Agreement is non levy AND EOI.

cofaulco: Ok I think this is fine actually as Non-levy users who have signed a Levy agreement should be able to see both links for finance

Stephek: Updated

PR #1771, opened 2019-07-29
AML-3763 About Your MoU

Howesy77:


Howesy77:
Content is confirmed as ok

src/SFA.DAS.EmployerAccounts.Web/Helpers/ControllerConstants.cs
```diff
@@ -2,7 +2,9 @@
 {
     public static class ControllerConstants
     {
-        public const string AboutYourAgreement = "AboutYourAgreement";
+        public const string AboutYourAgreementActionName = "AboutYourAgreement";
```
dashton82: its probably a bigger refactoring exercise, but splitting these into two classes for view names and actions, and not calling the property names xxxViewName, you could possibly do:
```
public static string AboutYourAgreement => nameof(AboutYourAgreement);
```

src/SFA.DAS.EmployerAccounts.Web/Views/EmployerAgreement/AboutYourMou.cshtml
```diff
@@ -0,0 +1,38 @@
+﻿@using SFA.DAS.Authorization
+@using SFA.DAS.Authorization.Mvc
+@using SFA.DAS.EmployerAccounts.Web.Helpers
+
+@model OrchestratorResponse<EmployerAgreementViewModel>
+
+@{
+    ViewBag.PageID = "mou-about";
+    ViewBag.Title = "About your MoU";
```
dashton82: is `MoU` correct to be shown in the title?

src/SFA.DAS.EmployerAccounts.Web/Helpers/ControllerConstants.cs
MichaelYoung1981: I agree, it's got to the size where we should split it out but not for now.  I'll do it the next time I'm in that class.

src/SFA.DAS.EmployerAccounts.Web/Views/EmployerAgreement/AboutYourMou.cshtml
MichaelYoung1981: I'm pretty sure it is...EoI, MoU...don't you love interchangeable terms.

Howesy77: It should probably be document - didn't have a design when this page was done :)

PR #1778, opened 2019-07-31
Aml 3775 add error to mou

MichaelYoung1981:


Winsto:
Should this also be [compared to base other than master](https://github.com/SkillsFundingAgency/das-employerapprenticeshipsservice/compare/AML-3574-legal-agreement...AML-3775_AddErrorToMoU) ?

I have just had a look at that diff and generally looks ok, but seems as though a whole new view is added when the user story is only talking about adding an error display on no options chosen

MichaelYoung1981:
> Should this also be [compared to base other than master](https://github.com/SkillsFundingAgency/das-employerapprenticeshipsservice/compare/AML-3574-legal-agreement...AML-3775_AddErrorToMoU) ?
> 
> I have just had a look at that diff and generally looks ok, but seems as though a whole new view is added when the user story is only talking about adding an error display on no options chosen

Sorry, meant to explain on the slack channel but forgot!  Slightly complicated, this is based on 3774 but also has changes from 3764 merged in!  Good luck getting git to work out that diff!

Howesy77:
We have merge conflicts

src/SFA.DAS.EmployerAccounts.Web/Controllers/EmployerAgreementController.cs
```diff
@@ -132,13 +132,20 @@ public async Task<ActionResult> SignAgreement(GetEmployerAgreementRequest reques
...
-            if (choice != ReviewAgreementLater)
+            if (choice == ReviewAgreementLater)
             {
                 return RedirectToAction(ControllerConstants.IndexActionName, ControllerConstants.EmployerTeamControllerName);
             }
 
             var userInfo = OwinWrapper.GetClaimValue(ControllerConstants.UserRefClaimKeyName);
             var agreement = await _orchestrator.GetById(agreementId, hashedAccountId, userInfo);
+
+            if (choice == null)
```
Howesy77: Would it be better to have choice default to 0 so a const can be used to define the state?

MichaelYoung1981: Isn’t the default value passed for a form field with no value null? I’m not sure zero would work and it’d be inconsistent with all other pages.

PR #1783, opened 2019-08-02
Aml 3768 all agreements eoi

Winsto:
The GitHub diff is showing spurious diffs between these two branches; I do not know why. I have been through and checked out files from the parent AML-3762-EOI-whitelist branch, but the diffs still show. If I diff across the branches locally then no diffs; no idea what GitHub is doing.

I a doing PR to AML-3762-EIO-whitelist whilst the EOI stuff is not yet test complete.


src/SFA.DAS.EmployerAccounts.Web.UnitTests/Controllers/EmployerAgreementControllerTests.cs
```diff
@@ -68,79 +68,7 @@ public Task ConfirmRemoveOrganisation_WhenIRemoveAlegalEntityFromAnAccount_ThenT
                     Assert.IsNotNull(actualModel);
                     Assert.IsNotNull(actualModel.FlashMessage);
                 });
-        }
```
Winsto: I do not know how to make this diff go away, it is nonsense. The file is checked out at parent branch state. Local Git diff shows no diffs.

src/SFA.DAS.EmployerAccounts.Web/Controllers/EmployerAgreementController.cs
```diff
@@ -18,12 +18,11 @@ namespace SFA.DAS.EmployerAccounts.Web.Controllers
     [RoutePrefix("accounts/{HashedAccountId}")]
     public class EmployerAgreementController : BaseController
     {
+        private const int ReviewAgreementLater = 2;
```
Winsto: Similarly to above, this diff came about from my merge conflict resolution. I have since checked the file out at parent branch state, but for some reason GitHub still insists it is a diff.

Winsto:
The implementation I have used is broken (at least on the branch) due to an exception when evaluating the results of the repository.

PR #1790, opened 2019-08-05
Set agreement type to EoI if an EoI agreement already exists.

MichaelYoung1981:


src/SFA.DAS.EmployerAccounts/Data/EmployerAgreementRepository.cs
```diff
@@ -151,5 +152,19 @@ public async Task<AccountLegalEntityModel> GetAccountLegalEntity(long accountLeg
 
             return result.SingleOrDefault();
         }
+
+        public async Task<IEnumerable<EmployerAgreement>> GetAccountAgreements(long accountId)
```
Howesy77: We seem to have various methods to access the DB should we adopt a consistent approach?

MichaelYoung1981:
> Do we need to add code / tests for a non EOI account that has an EOI user add an agreement?

Good question...I haven't considered that scenario as it isn't in the story.  I'll speak to Hardip and see if it is required.

PR #1793, opened 2019-08-06
Con 747 nuget update

Stephek:


cofaulco:
No longer required. The changes made here were addressed as part of the quality ticket to upgrade the nservicebus stuff. This was merged into the app service branch and deployed at the same time as that went live

PR #1795, opened 2019-08-06
Updated the message as per the story

tom-gough:


src/SFA.DAS.EmployerAccounts.Web/Controllers/EmployerAgreementController.cs
```diff
@@ -174,7 +174,7 @@ public async Task<ActionResult> Sign(string agreementId, string hashedAccountId,
                 else
                 {
                     flashMessage.Headline = "All agreements signed";
-                    flashMessage.Message = "You've successfully signed all of your organisation agreements.";
+                    flashMessage.Message = "You’ve successfully signed your organisation agreement(s)";
```
Winsto: As long as PO review says this changed apostrophe is ok :)

tom-gough: copied and pasted it from the story, agree it may not be intentional, will see if it looks acceptable at PO review.

PR #1797, opened 2019-08-07
Aml 3788 mou agreement confirmation message

tom-gough:


src/SFA.DAS.EmployerAccounts.Web/Controllers/EmployerAgreementController.cs
```diff
@@ -155,13 +155,19 @@ public async Task<ActionResult> Sign(string agreementId, string hashedAccountId,
...
 
-                if (response.Data.HasFurtherPendingAgreements)
+                if (agreement.Data.EmployerAgreement.AgreementType == AgreementType.NonLevyExpressionOfInterest)
+                {
+                    flashMessage.Headline = "Memorandum of Understanding signed";
+                    flashMessage.Message = "You’ve successfully signed the Memorandum of Understanding for your organisation.";
+                    result = RedirectToAction(ControllerConstants.IndexActionName, ControllerConstants.EmployerTeamControllerName);
+                }
+                else if (response.Data.HasFurtherPendingAgreements)
                 {
```
Winsto: What if there are pending memorandums of understanding to sign? Or is that not a thing?

PR #1798, opened 2019-08-08
Con 661 add training provider name on apprentice panel

Stephek:


src/SFA.DAS.EmployerAccounts.Web/Controllers/EmployerTeamController.cs
```diff
@@ -539,7 +540,21 @@ public ActionResult OtherTasksPanel(AccountDashboardViewModel model)
         [ChildActionOnly]
         public ActionResult ApprenticeshipDetails(AccountDashboardViewModel model)
         {
-            return PartialView(model);
+            Cohort cohort = model.AccountViewModel.Organisations.FirstOrDefault()?.Cohorts?.FirstOrDefault();
+            Apprenticeship apprenticeship = cohort?.Apprenticeships?.FirstOrDefault();
+            
+            var viewModel = new ApprenticeDetailsViewModel
```
cofaulco: Nice refactor!

PR #1800, opened 2019-08-08
Aml 3634 second look

Winsto:
- Bring back controller move
- fix issue with double encoding of '/' in PAYE ref
- bring up to date with master


Winsto:
Don't merge this yet, I think there is a problem.

This PR should also be introducing `SFA.DAS.EmployerAccount.Api.Types` to produce a nuget package that will replace the older references to `SFA.DAS.Account.Api.Types` project.

However that new project has been merged out now, so whilst this does not remove the old project there is no new project to put the new types in.

Winsto:
As per comment above.. need to re-instate project merged out by mistake.

PR #1803, opened 2019-08-09
CON-819 Use latest version of SFA.DAS.Http

lazcool:
Note: the latest version of `SFA.DAS.Http` doesn't contain `ApiClientBase`, which the commitments client requires, so this PR also switches out the commitments client for our own lightweight commitments service (in a similar style to the recruit service).


src/SFA.DAS.EAS.Portal/Application/Services/Commitments/CommitmentsService.cs
```diff
@@ -0,0 +1,30 @@
...
+        private readonly ILogger<CommitmentsService> _log;
+        private readonly RestHttpClient _restHttpClient;
+
+        public CommitmentsService(ICommitmentsApiHttpClientFactory commitmentsApiHttpClientFactory, ILogger<CommitmentsService> log)
+        {
+            _log = log;
+            _restHttpClient = new RestHttpClient(commitmentsApiHttpClientFactory.CreateHttpClient());
+        }
+
+        public Task<CommitmentView> GetProviderCommitment(long providerId, long commitmentId,
```
cofaulco: I like all params to be on separate lines if you have to split at all
`public Task<CommitmentView> GetProviderCommitment(
                                                                                                long providerId, 
                                                                                                long commitmentId, 
                                                                                                CancellationToken cancellationToken = default)`

src/SFA.DAS.EAS.Portal/Application/Services/Commitments/CommitmentsService.cs
```diff
@@ -0,0 +1,30 @@
...
+            _restHttpClient = new RestHttpClient(commitmentsApiHttpClientFactory.CreateHttpClient());
+        }
+
+        public Task<CommitmentView> GetProviderCommitment(long providerId, long commitmentId, 
+                CancellationToken cancellationToken = default)
+        {
+            _log.LogInformation($"Getting commitment {commitmentId} for provider {providerId}");
+
+            return _restHttpClient.Get<CommitmentView>($"/api/provider/{providerId}/commitments/{commitmentId}",
+        null, cancellationToken);
```
cofaulco: formatting

src/SFA.DAS.EAS.Portal/Application/Services/Commitments/Http/CommitmentsApiHttpClientFactory.cs
```diff
@@ -0,0 +1,33 @@
...
+        {
+            var httpClient = new HttpClientBuilder()
+                .WithDefaultHeaders()
+                .WithBearerAuthorisationHeader(new JwtBearerTokenGenerator(_commitmentsApiClientConfiguration))
+                .Build();
+
+            httpClient.BaseAddress = new Uri(_commitmentsApiClientConfiguration.BaseUrl);
+
+            //todo: set timeout
+            //httpClient.Timeout = TimeSpan.Parse(_recruitApiClientConfig.TimeoutTimeSpan);
```
cofaulco: if this isn't being done now do we need to make sure there is some tech debt or a story somewhere to address this later down the line ?

src/SFA.DAS.EAS.Portal/Application/Services/Commitments/ICommitmentsService.cs
```diff
@@ -0,0 +1,12 @@
+using System.Threading;
+using System.Threading.Tasks;
+using SFA.DAS.Commitments.Api.Types.Commitment;
+
+namespace SFA.DAS.EAS.Portal.Application.Services.Commitments
+{
+    public interface ICommitmentsService
+    {
+        Task<CommitmentView> GetProviderCommitment(long providerId, long commitmentId,
+            CancellationToken cancellationToken = default);
```
cofaulco: Again params on different lines

src/SFA.DAS.EAS.Portal/Application/Services/Commitments/Http/CommitmentsApiHttpClientFactory.cs
lazcool: It's not critical to have a timeout here, as it is when the client call API's, as the call is made in the context of the worker. I'll remove the comment for now, and we can always add a timeout if we really need it.

src/SFA.DAS.EAS.Portal/Application/Services/Commitments/CommitmentsService.cs
lazcool: I tend to split lines if they go past the 120 chars visual guide, but not too worried about formatting 😄. I've tidied up the formatting, putting params onto separate lines as you like.

src/SFA.DAS.EAS.Portal/Application/Services/Commitments/CommitmentsService.cs
lazcool: Tidied up.

src/SFA.DAS.EAS.Portal/Application/Services/Commitments/ICommitmentsService.cs
lazcool: Tidied.

PR #1804, opened 2019-08-09
Aml 3650 store account in finance db

MichaelYoung1981:


src/SFA.DAS.EmployerFinance/Data/AccountRepository.cs
```diff
@@ -26,25 +28,53 @@ public async Task<string> GetAccountName(long accountId)
...
         }
 
         public async Task<Dictionary<long, string>> GetAccountNames(IEnumerable<long> accountIds)
         {
-            var result = await _db.Value.Database.Connection.QueryAsync<AccountNameItem>(
+            var result = await _accountDb.Value.Database.Connection.QueryAsync<AccountNameItem>(
                 sql: "SELECT Id, Name FROM [employer_account].[Account] WHERE Id IN @accountIds",
                 param: new { accountIds = accountIds },
-                transaction: _db.Value.Database.CurrentTransaction.UnderlyingTransaction);
+                transaction: _accountDb.Value.Database.CurrentTransaction.UnderlyingTransaction);
```
tom-gough: is the idea that after the one time script is ran the _accountDb will be removed completely as a dependency from finance?

src/SFA.DAS.EAS.Web/Views/Shared/_Layout.cshtml
```diff
@@ -2,6 +2,7 @@
 @using SFA.DAS.Authorization
 @using SFA.DAS.Authorization.Mvc
 @using SFA.DAS.EAS.Web
+@using SFA.DAS.EAS.Web.Extensions
```
tom-gough: Is this needed/part of this story?

MichaelYoung1981: Nope...but it annoyed me so I changed it.  It caused prod error log entries so it should probably its own story,

src/SFA.DAS.EmployerFinance/Data/AccountRepository.cs
MichaelYoung1981: Yep, there is a story for the data migration.

PR #1805, opened 2019-08-12
AML-3642 Migrate api/user

lazcool:


src/SFA.DAS.EmployerAccounts.Api/Controllers/EmployerUserController.cs
```diff
@@ -0,0 +1,32 @@
...
+        {
+            _orchestrator = orchestrator;
+        }
+
+        [System.Web.Http.Route("accounts", Name = "Accounts")]
+        [ApiAuthorize(Roles = "ReadUserAccounts")]
+        [System.Web.Http.HttpGet]
+        public async Task<IHttpActionResult> GetUserAccounts(string userRef)
+        {
+            return Ok(await _orchestrator.GetUserAccounts(userRef));
```
lazcool: This code has been slightly simplified from the original. The original orchestrator method `GetUserAccounts` always returned a hard-coded `HttpStatusCode.OK` in an `OrchestratorResponse`, so the orchestrator method was changed to return the old data payload (`ICollection<AccountDetailViewModel>`), and this controller action was simplified to hard-code an `OK` response.

src/SFA.DAS.EmployerAccounts/Mappings/AccountMappings.cs
```diff
@@ -11,6 +12,21 @@ public AccountMappings()
...
             CreateMap<Account, AccountContext>();
             CreateMap<Account, AccountDto>();
+
+            CreateMap<Account, AccountDetailViewModel>()
+                .ForMember(target => target.AccountId, opt => opt.MapFrom(src => src.Id))
+                .ForMember(target => target.HashedAccountId, opt => opt.MapFrom(src => src.HashedId))
+                .ForMember(target => target.PublicHashedAccountId, opt => opt.MapFrom(src => src.PublicHashedId))
+                .ForMember(target => target.DateRegistered, opt => opt.MapFrom(src => src.CreatedDate))
+                .ForMember(target => target.DasAccountName, opt => opt.MapFrom(src => src.Name))
+                .ForMember(target => target.OwnerEmail, opt => opt.Ignore())
```
lazcool: The original mapping wasn't covered by an unit test calling `AssertConfigurationIsValid`, but this mapping profile _is_ covered, so we have to explicitly ignore fields that aren't mapped.

Note that the code, pre-move, was working with `Account` from the Domain project, whereas the already moved `GetUserAccountsQueryResponse` returns `Account` from `EmployerAccounts.Models.Account`.

It appears that `Account` has been forked from domain into the models in accounts, and has then diverged.
The domain `Account` has continued to be used by (and updated for) the API, whilst the model version has been used by (and updated for) accounts web.

MichaelYoung1981: We should really be returning a specific type from GetUserAccounts rather than a half populated object...I realise you've just copied this from the existing code base so I'm struggling to decide if we should tidy it up now or not.

lazcool: It has been agreed that only low-risk tidy-ups should be tackled in the first phase, so a [task](https://skillsfundingagency.atlassian.net/browse/AML-3809) has been created for this.

PR #1810, opened 2019-08-13
Aml 3633 move second action

Winsto:


src/SFA.DAS.EmployerAccounts.UnitTests/Queries/GetAccountPAYESchemes/WhenIGetAccountPayeSchemes.cs
```diff
@@ -13,7 +14,7 @@
 
 namespace SFA.DAS.EmployerAccounts.UnitTests.Queries.GetAccountPAYESchemes
 {
-    class WhenIGetAccountPayeSchemes : QueryBaseTest<GetAccountPayeSchemesQueryHandler, GetAccountPayeSchemesQuery, GetAccountPayeSchemesResponse>
+    class WhenIGetAccountPayeSchemes : QueryBaseTest<GetAccountPayeSchemesForAuthorisedUserQueryHandler, GetAccountPayeSchemesForAuthorisedUserQuery, GetAccountPayeSchemesResponse>
```
MichaelYoung1981: Tests should be specific to the class under test rather than testing 2 at once.

Winsto: As per discussion; common code extracted into new service `PayeSchemesService`

Query handlers both updated to call this service.

No update of test styles or scenarios, but scenarios covering expected results for original command handler now applied to service.

src/SFA.DAS.EmployerAccounts.UnitTests/Queries/GetAccountPAYESchemes/WhenIGetAccountPayeSchemes.cs
```diff
@@ -48,31 +46,25 @@ public void Arrange()
                 Amount = 0.5m
             };
 
+            _payeView.EnglishFraction = _englishFraction;
```
MichaelYoung1981: _englishFraction isn't needed in the test any more as you're just doing an object comparison with whatever has been returned from the new service

src/SFA.DAS.EmployerAccounts.UnitTests/Queries/GetAccountPAYESchemesForAuthorisedUser/WhenIGetAccountPayeSchemes.cs
```diff
@@ -0,0 +1,112 @@
...
+            };
+
+            _englishFraction = new DasEnglishFraction
+            {
+                EmpRef = _payeView.Ref,
+                DateCalculated = UpdateDate,
+                Amount = 0.5m
+            };
+
+            _payeView.EnglishFraction = _englishFraction;
```
MichaelYoung1981: _englishFraction isn't needed in the test any more as you're just doing an object comparison with whatever has been returned from the new service

src/SFA.DAS.EmployerAccounts.UnitTests/Services/PayeSchemes/WhenIGetAccountPayeSchemes.cs
```diff
@@ -0,0 +1,121 @@
...
+            _payeSchemesRepository = new Mock<IPayeRepository>();
+            _englishFractionsRepository = new Mock<IEnglishFractionRepository>();
+            _hashingService = new Mock<IHashingService>();
+
+            _payeSchemesRepository.Setup(x => x.GetPayeSchemesByAccountId(It.IsAny<long>())).ReturnsAsync(new List<PayeView>
+            {
+                _payeView
+            });
+
+            _englishFractionsRepository.Setup(x => x.GetCurrentFractionForSchemes(It.IsAny<long>(), It.IsAny<IEnumerable<string>>()))
```
MichaelYoung1981: Are we able to remove the IsAnys and replace with specific values?

src/SFA.DAS.EmployerAccounts/Queries/GetAccountPayeSchemes/GetAccountPayeSchemesQueryHandler.cs
```diff
@@ -41,24 +36,11 @@ public async Task<GetAccountPayeSchemesResponse> Handle(GetAccountPayeSchemesQue
             {
                 throw new UnauthorizedAccessException();
```
MichaelYoung1981: Isn't this code now redundant as the validator no longer checks authorisation?

src/SFA.DAS.EmployerAccounts/Services/PayeSchemesService.cs
```diff
@@ -0,0 +1,66 @@
...
+            IPayeRepository payeRepository,
+            IEnglishFractionRepository englishFractionRepository,
+            IHashingService hashingService
+            )
+        {
+            _payeRepository = payeRepository;
+            _englishFractionRepository = englishFractionRepository;
+            _hashingService = hashingService;
+        }
+        public async Task<IEnumerable<PayeView>> GetPayeSchemsWithEnglishFractionForHashedAccountId(string hashedAccountId)
```
MichaelYoung1981: The method looks like it could do with some tidy up refactoring.

src/SFA.DAS.EmployerAccounts.UnitTests/Queries/GetAccountPAYESchemes/WhenIGetAccountPayeSchemes.cs
Winsto: I just left it as it was part of the object being compared, but not necessary agreed. Removed.

src/SFA.DAS.EmployerAccounts.UnitTests/Queries/GetAccountPAYESchemesForAuthorisedUser/WhenIGetAccountPayeSchemes.cs
Winsto: I just left it as it was part of the object being compared, but not necessary agreed. Removed.

src/SFA.DAS.EmployerAccounts.UnitTests/Services/PayeSchemes/WhenIGetAccountPayeSchemes.cs
Winsto: Yes, I have just left these tests as they were have not looked at re-writing them. I am happy to look at writing better tests, just used what was already there.

src/SFA.DAS.EmployerAccounts/Queries/GetAccountPayeSchemes/GetAccountPayeSchemesQueryHandler.cs
Winsto: It is if I remove the test that covers it as well, which I have done.

However I have a bit of an issue with this; not with your comment, but in the whole approach.

So I could argue that the line should stay in, because the Handler doesn't know whether the Validator might return unauthorized or not, that is not a 'concern' of the Handler. A `ValidationResult` could have `IsUnauthorized` set.

Thinking about it my issue is more that we have this same code in all the handlers. It seems like the validation should be handled outside of the handlers or maybe the validators should be throwing their own exception.

src/SFA.DAS.EmployerAccounts/Services/PayeSchemesService.cs
Winsto: Just copied and pasted, will take a close look.

Winsto: Done quick extract method... more could be done in that section dealing with English Fraction...

depends how much we want to change as part of this..let me know.

DannyGB:
Original author is no-longer here and the PR is over 9 months old, it will never be merged

PR #1811, opened 2019-08-13
AML-3635 Move Transfer Connections API

Howesy77:


src/SFA.DAS.EAS.Account.Api.UnitTests/SFA.DAS.EAS.Account.Api.UnitTests.csproj
```diff
@@ -1,275 +1,274 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
```
Winsto: Why is it showing the whole project file as a diff?

PR #1812, opened 2019-08-13
Aml 3653 add paye schemes to finance db

MichaelYoung1981:


src/SFA.DAS.EmployerFinance.MessageHandlers.UnitTests/CommandHandlers/GivenAPayeSchemeIsAddedToAnAccount.cs
```diff
@@ -0,0 +1,69 @@
...
+            var name = "Scheme Name";
+            var empRef = "ABC/123246";
+            var aorn = "AORN";
+
+            await _handler.Handle(new CreateAccountPayeCommand(accountId, empRef, name, aorn), _context.Object);
+
+            _context.Verify(x => x.Send(It.IsAny<ImportAccountLevyDeclarationsCommand>(), It.IsAny<SendOptions>()), Times.Never());
+        }
+    }
+}
```
Winsto: Is it worth having a scenario to specify that if we cannot save the paye scheme then we do not attempt to import levy declarations?

That said, if importing levy declarations would cause no harm in the event of no existing payescheme then it wouldn't matter.

Winsto:
Looked through updates and discussed.

Moot point about test coverage, but original approval still stands.

PR #1819, opened 2019-08-15
Aml 3643 move accounts legalentities

Winsto:
Note: I have had to add the route for `GetLegelEntity` (singular) in the new controller in order to return the resource list from the multiple get method (should really have moved the methods the other way around). If this looks too ugly I can move the singular action as part of this PR as well, then it is all tied up in one PR/branch.


src/SFA.DAS.EmployerAccounts/Queries/GetAccountLegalEntitiesByHashedAccountId/GetAccountLegalEntitiesByHashedAccountIdResponse.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System.Collections.Generic;
+using SFA.DAS.EmployerAccounts.Models.Account;
+
+namespace SFA.DAS.EmployerAccounts.Queries.GetAccountLegalEntitiesByHashedAccountId
+{
+    public class GetAccountLegalEntitiesByHashedAccountIdResponse
+    {
+        public List<AccountSpecificLegalEntity> Entities { get; set; }
```
MichaelYoung1981: Can this be renamed to LegalEntities?

Winsto: It can and just has.

I had already changed it from `Entites` :0)

I've updated the same property name on the original `GetLegalEntitesResponse` as well.

Winsto:
Just found some integration tests that should have been sorted out as part of this, sorry.

Winsto:
As per discussion, separate ticket raised to cover moving of integration tests; as some other api moves have not had associated integration tests move as well. So will reopen this then we can test.

MichaelYoung1981:
The integration tests are still failing...we can't commit broken tests.

PR #1824, opened 2019-08-20
Aml 3652 remove legal entities in finance

tom-gough:


src/SFA.DAS.EmployerFinance.MessageHandlers.UnitTests/EventHandlers/RemovedLegalEntityEventHandlerTests.cs
```diff
@@ -0,0 +1,66 @@
...
+using SFA.DAS.EmployerAccounts.Messages.Events;
+using SFA.DAS.EmployerFinance.Commands.RemoveAccountLegalEntity;
+using SFA.DAS.EmployerFinance.MessageHandlers.EventHandlers;
+using SFA.DAS.Testing;
+
+namespace SFA.DAS.EmployerFinance.MessageHandlers.UnitTests.EventHandlers
+{
+    [TestFixture]
+    [Parallelizable]
+    internal class RemovedLegalEntityEventHandlerTests : FluentTest<RemovedLegalEntityEventHandlerTestsFixture>
```
MichaelYoung1981: Should that be internal?

src/SFA.DAS.EAS.Web/Views/Shared/_Layout.cshtml
```diff
@@ -2,6 +2,7 @@
 @using SFA.DAS.Authorization
 @using SFA.DAS.Authorization.Mvc
 @using SFA.DAS.EAS.Web
+@using SFA.DAS.EAS.Web.Extensions
```
Winsto: Is this really part of this change?

MichaelYoung1981: No, I made the change as part of a branch this is based on because it was causing thousands of prod errors and it annoyed me.

PR #1839, opened 2019-08-27
Changed to only save cookie once as multiple calls were replacing pre…

MichaelYoung1981:
…vious updates.

### DIFF
--- a/src/SFA.DAS.CloudService/SFA.DAS.EAS.WebContent/diagnostics.wadcfgx
+++ b/src/SFA.DAS.CloudService/SFA.DAS.EAS.WebContent/diagnostics.wadcfgx
@@ -43,5 +43,5 @@
   <PrivateConfig>
     <StorageAccount name="" key="" endpoint="" />
   </PrivateConfig>
-  <IsEnabled>true</IsEnabled>
+  <IsEnabled>false</IsEnabled>
 </DiagnosticsConfiguration>
--- a/src/SFA.DAS.EmployerAccounts.UnitTests/Commands/CookieBasedOrganisationAndPayeRefDataSaver/Given_Cookie_Does_Not_Already_Exist.cs
+++ b/src/SFA.DAS.EmployerAccounts.UnitTests/Commands/CookieBasedOrganisationAndPayeRefDataSaver/Given_Cookie_Does_Not_Already_Exist.cs
@@ -1,0 +1,71 @@
+﻿using System.Threading.Tasks;
+using Moq;
+using NUnit.Framework;
+using SFA.DAS.EmployerAccounts.Commands.OrganisationAndPayeRefData;
+using SFA.DAS.EmployerAccounts.Interfaces;
+using SFA.DAS.EmployerAccounts.Models.Account;
+using SFA.DAS.Testing;
+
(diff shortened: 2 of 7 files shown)


src/SFA.DAS.CloudService/SFA.DAS.EAS.WebContent/diagnostics.wadcfgx
```diff
@@ -43,5 +43,5 @@
   <PrivateConfig>
     <StorageAccount name="" key="" endpoint="" />
   </PrivateConfig>
-  <IsEnabled>true</IsEnabled>
+  <IsEnabled>false</IsEnabled>
```
tom-gough: what is the motivation behind this change?

MichaelYoung1981: Dev Ops removed a load of the diagnostics config so the app wouldn't run locally without making that change.

src/SFA.DAS.EmployerAccounts.UnitTests/Commands/CookieBasedOrganisationAndPayeRefDataSaver/Given_Cookie_Does_Not_Already_Exist.cs
```diff
@@ -0,0 +1,71 @@
...
+                        act: f => f.Handle(),
+                        assert: f =>
+                            f.CookieRepository
+                                .Verify(
+                                    m =>
+                                        m.Update(
+                                            It.IsAny<string>(),
+                                            It.IsAny<EmployerAccountData>()),
+                                    Times.Never));
+            }
```
tom-gough: indentation looks slightly off here

src/SFA.DAS.EmployerAccounts/Commands/OrganisationAndPayeRefData/CookieBasedOrganisationAndPayeDataSaver.cs
```diff
@@ -0,0 +1,64 @@
+﻿using System;
+using System.Threading.Tasks;
+using MediatR;
+using SFA.DAS.EmployerAccounts.Interfaces;
+using SFA.DAS.EmployerAccounts.Models.Account;
+
+namespace SFA.DAS.EmployerAccounts.Commands.OrganisationAndPayeRefData
+{
+    public sealed  class CookieBasedOrganisationAndPayeDataSaver : AsyncRequestHandler<SaveOrganisationAndPayeData>
```
tom-gough: why not handler?

src/SFA.DAS.CloudService/SFA.DAS.EAS.WebContent/diagnostics.wadcfgx
tom-gough: what is the motivation behind this change?

src/SFA.DAS.EmployerAccounts/Commands/OrganisationAndPayeRefData/CookieBasedOrganisationAndPayeDataSaver.cs
MichaelYoung1981: I was following the existing naming convention of the command I'd copied.  I did think about changing it but it would have then been inconsistent.

PR #1840, opened 2019-08-27
Aml 3609 move sfa das binders

tom-gough:


Winsto:
Looking at latest master it seems we can achieve this by just removing the `TrimStringModelBinder` in `SFA.DAS` without updating the testing package references.

Which might mean we do not need the trim string in the shared testing package (although removing that now might be awkward; don't know that someone else is not using it).

There is no harm in updating the package reference (especially if we decide to leave the trim string binder in there), so can approve if that's the way we are going. Can talk about it.

PR #1841, opened 2019-08-27
Aml 3644 move accounts getlegalentity

Winsto:
I have made PR against parent branch rather than [base api migration branch](https://github.com/SkillsFundingAgency/das-employerapprenticeshipsservice/tree/AML-3634-second-look) as these keeps diffs to a minimum and this work builds on the work of moving the multiple legal entities api lookup.


src/SFA.DAS.EmployerApprenticeshipsService.Application/SFA.DAS.EAS.Application.csproj
```diff
@@ -383,8 +383,6 @@
     <Compile Include="Mappings\LegalEntityMappings.cs" />
     <Compile Include="Queries\Extensions\QueryExtensions.cs" />
     <Compile Include="Queries\GetLegalEntity\GetLegalEntityQuery.cs" />
```
Howesy77: Can the query be removed from this project?

Winsto: Yes it could; I left it as I was just doing C&P, but could turn it into primitive parameters is we want to do that as well.

Winsto: Updated to primitive parameters.

PR #1844, opened 2019-08-29
AML-3690: updated contents as per specificaion in story. Note: Find l…

Winsto:
…ink not re-introduced.


Winsto:
The content changes here will now be done as part of [the other](https://github.com/SkillsFundingAgency/das-employerapprenticeshipsservice/pull/1846) change.

PR #1851, opened 2019-09-03
Aml 3877 fix hashedaccountid propagation

tom-gough:


src/SFA.DAS.EmployerAccounts/Commands/AddPayeToAccount/AddPayeToAccountCommandValidator.cs
```diff
@@ -47,10 +47,6 @@ public async Task<ValidationResult> ValidateAsync(AddPayeToAccountCommand item)
 
         private static void CheckFieldsArePopulated(AddPayeToAccountCommand item, ValidationResult validationResult)
         {
-            if (string.IsNullOrEmpty(item.AccessToken))
```
CodeWolf86: What was the need to remove this validation? Is this down to the fact sometimes we are arriving via hashed account id and so the access token isn't present? but we will now always have the hashed account id regardless?

src/SFA.DAS.EmployerAccounts.Web/Controllers/SearchPensionRegulatorController.cs
```diff
@@ -99,9 +103,19 @@ public async Task<ActionResult> SearchPensionRegulator(string hashedAccountId, S
         }
 
         [HttpGet]
-        [Route("pensionregulator/aorn")]
-        public async Task<ActionResult> SearchPensionRegulatorByAorn(string payeRef, string aorn)
+        [Route("{HashedAccountId}/pensionregulator/aorn", Order = 0)]
+        [Route("pensionregulator/aorn", Order = 1)]
+        public async Task<ActionResult> SearchPensionRegulatorByAorn(string payeRef, string aorn, string hashedAccountId)
         {
+            if (!string.IsNullOrWhiteSpace(hashedAccountId))
```
CodeWolf86: So is the cookie a removal of the same cookie which is then re-added, or are we removing a cookie which we want to overwrite with the hashed account id? ( Appreciate this is mimicking the GovernmentGateway signin page, just trying to understand it)

PR #1857, opened 2019-09-10
Con 820 better handle account info

DannyGB:


src/SFA.DAS.EmployerAccounts.Web/Orchestrators/EmployerTeamOrchestrator.cs
```diff
@@ -190,7 +191,7 @@ public async Task<OrchestratorResponse<AccountDashboardViewModel>> GetAccount(st
                 var pendingAgreements = agreementsResponse.EmployerAgreements.Where(a => a.HasPendingAgreement).Select(a => new PendingAgreementsViewModel { HashedAgreementId = a.Pending.HashedAgreementId }).ToList();
                 var tasks = tasksResponse?.Tasks.Where(t => t.ItemsDueCount > 0 && t.Type != "AgreementToSign").ToList() ?? new List<AccountTask>();
                 var showWizard = userResponse.User.ShowWizard && userRoleResponse.UserRole == Role.Owner;
-                var accountDetailViewModel = apiGetAccountTask.Result;
+                var accountDetailViewModel = apiGetAccountTask;
```
cofaulco: no need to re-assign here

src/SFA.DAS.EmployerAccounts.Web/Orchestrators/EmployerTeamOrchestrator.cs
```diff
@@ -144,8 +145,8 @@ public async Task<OrchestratorResponse<EmployerTeamMembersViewModel>> ChangeRole
         public async Task<OrchestratorResponse<AccountDashboardViewModel>> GetAccount(string accountId, string externalUserId)
         {
             try
-            {
-                var apiGetAccountTask = _accountApiClient.GetAccount(accountId);
+            {
+                var apiGetAccountTask = await _accountApiClient.GetAccount(accountId).ConfigureAwait(false);
```
cofaulco: do we need to do this before firing off the other tasks? This should be an unlikely scenario so would the benefits of parallelising most of the time outweigh the edge case where we are doing unnecessary work ?

PR #1862, opened 2019-09-12
Agreement signed auth

tom-gough:


src/SFA.DAS.EmployerAccounts.Jobs/StartupJobs/CreateReadStoreDatabaseJob.cs
```diff
@@ -28,25 +29,54 @@ public async Task Run()
                 Id = DocumentSettings.DatabaseName
             };
 
-            var documentCollection = new DocumentCollection
+            var documentCollections = new List<DocumentCollection>
```
MichaelYoung1981: Will this job update the read store if it already exists?

src/SFA.DAS.EmployerAccounts.MessageHandlers/EventHandlers/SignedAgreementEventHandler.cs
```diff
@@ -25,6 +31,16 @@ public async Task Handle(SignedAgreementEvent message, IMessageHandlerContext co
...
                 message.UserName,
                 message.UserRef.ToString()));
+
+            try
+            {
+                await _mediator.Send(new SignAccountAgreementCommand(message.AccountId, message.AgreementVersion, message.AgreementType.ToString()));
+            }
+            catch (Exception e)
+            {
+                Console.WriteLine(e);
```
MichaelYoung1981: Shouldn't the exception be logged to Kibana?

tom-gough: yes, this is just some local debug code I left in, will remove

src/SFA.DAS.EmployerAccounts.Jobs/StartupJobs/CreateReadStoreDatabaseJob.cs
tom-gough: If you are talking about the scenario where the readstore exists with the existing table but not the new "AccountSignedAgreement" table then it will be fine as it does a create if not exists at the level of each table. One thing we need to consider is migrating existing signed agreements into this readstore when we deploy it for the first time.

tom-gough: solution for migrating existing agreements seems to be a run once job. I will write this next.

src/SFA.DAS.EmployerAccounts.Jobs/RunOnceJobs/SeedAccountSignedAgreementsJob.cs
```diff
@@ -0,0 +1,65 @@
...
+
+                if (!agreementExists)
+                {
+                    var document = new AccountSignedAgreement(agreement.AccountId, agreement.Template.VersionNumber, agreement.Template.AgreementType.ToString(), Guid.NewGuid());
+                    await _accountSignedAgreementsRepository.Add(document, null, CancellationToken.None);
+                }
+            }
+
+            _logger.LogInformation("Finished migrating signed agreements into the read store");
+        }
```
CodeWolf86: So this is something a little off the cuff to change the approach. It would need testing but illustrates the idea of rather than retrieving then looping over 1 by 1, we retrieve the data sets to do a local join and then can create the ones that are needed. 

Something to consider if we think the data set will be too large as we go through the environments.

```
public async Task MigrateAgreements()
        {
            var agreements = _db.Value.Agreements
                .ProjectTo<AgreementDto>(_configurationProvider)
                .Where(x => x.SignedDate != null).ToList();

            _logger.LogInformation("Migrating signed agreements into the read store");

            var financeAgreement = await CosmosDb.QueryableExtensions.ToListAsync(_accountSignedAgreementsRepository.CreateQuery());

            var existingAgreements = from a in agreements
                                     join f in financeAgreement
                                     on a.AccountId equals f.AccountId
                                     where a.Template.AgreementType.ToString() == f.AgreementType && a.Template.VersionNumber == f.AgreementVersion
                                     select a.Id;

            var agreementsToCreate = agreements.Where(s => !existingAgreements.Contains(s.Id));
            foreach (var agreement in agreementsToCreate)
            {
                var document = new AccountSignedAgreement(agreement.AccountId, agreement.Template.VersionNumber, agreement.Template.AgreementType.ToString(), Guid.NewGuid());
                await _accountSignedAgreementsRepository.Add(document, null, CancellationToken.None);
            }

            _logger.LogInformation("Finished migrating signed agreements into the read store");
        }
```

src/SFA.DAS.EmployerAccounts.Jobs/RunOnceJobs/SeedAccountSignedAgreementsJob.cs
```diff
@@ -46,11 +46,12 @@ public async Task MigrateAgreements()
 
             _logger.LogInformation("Migrating signed agreements into the read store");
 
+            var readStoreAgreements = await CosmosDb.QueryableExtensions.ToListAsync(
+                _accountSignedAgreementsRepository.CreateQuery());
+
             foreach (var agreement in agreements)
```
CodeWolf86: What do you think about this instead?

            // For All Agreements in the Employer DB
            // Find me all the agreements where
            // There is not any element in the financial agreement collection that matches AccountId Version and Agreement Type

            var differences = agreements.
                Where(agreement => !readStoreAgreements.Any
                    (fAgreement => fAgreement.AccountId == agreement.AccountId && fAgreement.AgreementVersion == agreement.Template.VersionNumber && fAgreement.AgreementType == agreement.Template.AgreementType.ToString()));

            // Depending on your preference for readability
            // The Inversion is
            // For All Agreements in the Employer DB
            // Find me all the agreements Where 
            // There is no financial agreement that matches accountID, Version and AgreementType in the financial agreement collection.

            var differences2 = agreements.
                Where(agreement => readStoreAgreements.
                All(fAgreement => fAgreement.AccountId != agreement.AccountId || fAgreement.AgreementVersion != agreement.Template.VersionNumber || fAgreement.AgreementType != agreement.Template.AgreementType.ToString()));


            foreach(var difference in differences)
            {
                var document = new AccountSignedAgreement(difference.AccountId, difference.Template.VersionNumber, difference.Template.AgreementType.ToString(), Guid.NewGuid());
                await _accountSignedAgreementsRepository.Add(document, null, CancellationToken.None);
            }

PR #1863, opened 2019-09-12
Con 730 navigate to finance

cofaulco:


src/SFA.DAS.EmployerFinance.Web/Views/EmployerAccountTransactions/Index.cshtml
```diff
@@ -18,113 +18,116 @@
...
+                var estimatedSpending = Model.Data.ProjectedSpend.ToString("C0", culture);
+                var estimatedSpendingClass = string.Empty;
+
+                if (estimatedSpending.Length > 8)
+                {
+                    estimatedSpendingClass = "long";
+                }
+                if (estimatedSpending.Length > 10)
+                {
+                    estimatedSpendingClass = "longer";
```
DannyGB: Is there a better way to do this, feels like a hack?

PR #1870, opened 2019-09-18
AML-3616 Move HTTP and HMRC Service

Howesy77:
Also includes AML-3610 Move SFA.Caches

Has been tested on AT. Potentially needs testing with production settings against the HRMC live service?


src/SFA.DAS.EmployerFinance/Http/HttpClientWrapper.cs
```diff
@@ -8,7 +8,7 @@
 using Newtonsoft.Json;
 using SFA.DAS.NLog.Logger;
 
-namespace SFA.DAS.Http
+namespace SFA.DAS.EmployerFinance.Http
```
CodeWolf86: How come the `HttpClientWrapper` has been moved to finance instead of making use of the das-shared/SFA.Das.Http package? Think you mentioned this as something you didn't want to do but had to do for simplicity?

CodeWolf86:
Merged master back into the branch to resolve conflicts after the api integration test merge.

src/SFA.DAS.EmployerFinance/Http/HttpClientWrapper.cs
Howesy77: There's no correlation between the HTTP client in SFA.Das.Http and our solution. I think in time the Finance project should be changed to work with the existing nuget package. But I don't think this is work for now.

PR #1871, opened 2019-09-19
Con 927 non levy co investment transactions

cofaulco:


src/SFA.DAS.EmployerFinance.Web.UnitTests/Orchestrators/WhenGettingCoursePaymentSummary.cs
```diff
@@ -0,0 +1,194 @@
...
+        [TestCase(1, Description = "Course payment summary for single apprentice")]
+        [TestCase(9, Description = "Course payment summary for multiple apprentice")]
+        public async Task ThenSummariesForEachApprenticeShouldBeCreated(int numberOfApprentices)
+        {
+            // Arrange
+            var coursePayments = CreateCoursePayments(numberOfApprentices, 1);
+            SetupGetCoursePaymentsResponse(2019, 9, coursePayments);
+
+            // Act
+            var response = await _sut.GetCoursePaymentSummary("abc123", 888888, "A course", 4, null, new DateTime(2019, 9, 1), new DateTime(2019, 9, 30), "userId");
```
DannyGB: Will the hard coded dates in the tests cause test failures when the year changes?

src/SFA.DAS.EmployerFinance.Web/Filters/LevyEmployerTypeOnly.cs
```diff
@@ -0,0 +1,43 @@
...
+
+namespace SFA.DAS.EmployerFinance.Web.Filters
+{
+    public class LevyEmployerTypeOnly : ActionFilterAttribute
+    {
+        public override void OnActionExecuting(ActionExecutingContext filterContext)
+        {
+            try
+            {
+                var hashedAccountId = filterContext.ActionParameters["HashedAccountId"].ToString();
```
DannyGB: Would it be better to check for HashedAccountId explicitly rather than rely on the try, catch?

src/SFA.DAS.EmployerFinance.Web/Orchestrators/EmployerAccountTransactionsOrchestrator.cs
```diff
@@ -152,21 +141,29 @@ public virtual async Task<OrchestratorResponse<FinanceDashboardViewModel>> Index
...
+                var getProviderPaymentsTask = _mediator.SendAsync(new FindAccountProviderPaymentsQuery
+                {
+                    HashedAccountId = hashedAccountId,
                     UkPrn = ukprn,
                     FromDate = fromDate,
                     ToDate = toDate,
                     ExternalUserId = externalUserId,
                 });
 
+                var providerPaymentsResponse = await getProviderPaymentsTask;
```
DannyGB: Would it make sense to parallelise these calls?

src/SFA.DAS.EmployerFinance.Web.UnitTests/Orchestrators/WhenGettingCoursePaymentSummary.cs
cofaulco: No issue with the dates, this is getting the summary for a timespan so doesn't matter

PR #1876, opened 2019-09-23
Prevent expiry of funds by changing the date.  Needs to be changed ba…

MichaelYoung1981:
…ck after 28th but before 10th.


satya-sanikommu:
This is just changing the Cron time, so that we dont run the Expiry of funds this month. It will be reverted as soon as move into October

PR #1877, opened 2019-09-24
Removing CloudConfigurationManager resolving of Configuration.

CodeWolf86:
Removing CloudConfigurationManager resolving of Configuration. 

This requires environment variables setup in the VSTS pipeline which DevOps are trying to move away from. As these were not set we kept getting 401 Unauthorised errors when accessing the new Accounts Api because it had no config to know where to validate its tokens against.

This will require an update to the Accounts Template Schema Json to make sure it's mapped across. ( I think )


CodeWolf86:
Fix won't work. Back to the drawing board.

PR #1882, opened 2019-09-26
Con 937 add paye scheme

Stephek:


src/SFA.DAS.EmployerAccounts.Web/Views/Shared/_Layout.cshtml
```diff
@@ -44,6 +44,7 @@
     <!--[if IE 8]><link href="@Styles.Url("~/bundles/screenie8")" media="screen" rel="stylesheet" /><![endif]-->
     <!--[if gt IE 8]><!-->
     <link href="@Styles.Url("~/bundles/screen")" media="all" rel="stylesheet" />
+    <link href="@Html.CdnLink("css", "main.css")" rel="stylesheet" />
```
cofaulco: We don't want to pull the styles from the CDN, work is being done to copy the styles required into the local stylesheet for now.

Stephek:
> 1 change regarding pulling the cdn into the V1 layout. Only other thing is, looking at the code, moving everything out of the EmployerTeam Index.cshtml makes for a little more confusing rendering the Rows and Panels which aren't so much a thing in this design. Could we have just called the ChildAction for rendering the call to action panel at the top of the existing view?. Just a thought, not saying we revert it all back

Stephek:
I'm assuming that the page will morph towards the layout of the new homepage so separated the layout from the content to more easily allow for this going forward.

src/SFA.DAS.EmployerAccounts.Web/Views/Shared/_Layout.cshtml
Stephek: Should I remove?  This story will be blocked until that work is complete.

cofaulco:
Feature toggled call to action for the 4 panel homepage which is no longer applicable. Closing PR and deleting branch

PR #1891, opened 2019-10-03
Feature call to action

Stephek:
Contains branches 

CON-937 add PAYE
CON-939 sign agreement
CON-940 reserve funding


src/SFA.DAS.EmployerAccounts/Configuration/ConfigurationKeys.cs
```diff
@@ -12,5 +12,6 @@ public static class ConfigurationKeys
         public const string EmployerFinance = "SFA.DAS.EmployerFinance";
         public const string EmployerAccountsReadStore = "SFA.DAS.EmployerAccounts.ReadStore";
         public const string AuditApi = "SFA.DAS.AuditApiClient";
+        public static string ReservationsClientApiConfiguration => $"SFA.DAS.Reservations.Api.Client";
```
cofaulco: should all these Configuration Keys be consistent ? either as const or static properties

Stephek: Yes , good point.  Changed

Stephek:
Using the child action means a unit test can be written to determine the view that is displayed in that area of the page.  The layout is two column and multiple rows which toggle on and off depending on the state of the viewmodel so the implementation models that to some extent.  The alternative is logic in the view which results in mixing style and logic.  Ideally we should be moving towards html web components which the child action/controller/separate view is closer to.
If you want the current implementation to be changed then please specify how it should be implemented and I can change it to match.

cofaulco:
If I were to change it I would still implement the call to action panel as a ChildAction but otherwise leave the rest of the Index.cshtml as was. My concern is that devs outside of this team will be adding, removing from this view and Row1Panel2 doesn't tell me where the Activities and Tasks list is. I agree that the ChildAction encapsulating the logic for what to display is better and can be tested. However when the toggle is off, which at this point seems like it may well be forever until if eventually gets removed altogether, there is logically only a single row with the 2 columns

PR #1900, opened 2019-10-15
CDN Changes to Registration Journey

CodeWolf86:


src/SFA.DAS.EmployerAccounts.Web/Views/Shared/_Layout_v2.cshtml
```diff
@@ -32,6 +32,8 @@
         document.body.className = ((document.body.className) ? document.body.className + ' js-enabled' : 'js-enabled');
     </script>
 
+    <br />THIS IS TEMPLAE _LAYOUT V2 *************<br />
```
jamesmking: This needs to come out

src/SFA.DAS.EAS.Portal/Startup/ConfigurationStartup.cs
```diff
@@ -15,7 +15,7 @@ public static IHostBuilder ConfigureDasAppConfiguration(this IHostBuilder builde
                 .AddEnvironmentVariables()
                 .AddAzureTableStorage(o =>
                     {
-                        o.EnvironmentNameEnvironmentVariableName = EnvironmentVariableName.EnvironmentName;
+                        o.EnvironmentNameEnvironmentVariableName = "LOCAL"; //EnvironmentVariableName.EnvironmentName;
```
jamesmking: Think this change needs reverting

PR #1904, opened 2019-10-22
Tweaked signed agreement queries to improve performance

MichaelYoung1981:


src/SFA.DAS.EmployerAccounts/Queries/GetEmployerAgreement/GetEmployerAgreementQueryHandler.cs
```diff
@@ -94,5 +90,15 @@ public async Task<GetEmployerAgreementResponse> Handle(GetEmployerAgreementReque
                 LastSignedAgreement = lastSignedAgreement
             };
         }
+
+        private string GetUserFullName(string userId)
+        {
+            var externalUserId = Guid.Parse(userId);
+            var user = _database.Value.Users
+                .Where(m => m.Ref == externalUserId)
+                .Select(m => m)
```
roysbailey: Why select the whole user, and then just return the name?  Could you not just select the name only?

MichaelYoung1981: FullName is actually a combination of FirstName and LastName so EF can't defer that to the database.

PR #1914, opened 2019-11-01
HotFix to resolve Import Levy Command pointing at wrong DBSchema for …

CodeWolf86:
…GetPayeSchemes Store Procedure


CodeWolf86:
> Approved, but in my opinion we shouldn't merge it to master without testing.

Understood. If we had a quicker build and deploy pipeline i would advocate sticking to usual procedures. I'm being informed that as a P2 incident it has to be fixed today. Due to the constraints of build, deploy , test, our pipeline puts at risk that SLA

PR #1917, opened 2019-11-04
CON-987_add new button

shomavg:


src/SFA.DAS.EAS.Support.Infrastructure/Settings/EmployerAccountsConfiguration.cs
```diff
@@ -0,0 +1,12 @@
+﻿using SFA.DAS.Messaging.AzureServiceBus.StructureMap;
+
+namespace SFA.DAS.EAS.Support.Infrastructure.Settings
+{
+    public class EmployerAccountsConfiguration : ITopicMessagePublisherConfiguration
```
cofaulco: do we need to implement the ITopicMessagePublisherConfiguration interface here? I think we only need the EmployerAccountsBaseUrl?

PR #1921, opened 2019-11-06
Qual 25 accounts api get account performance improvements

MichaelYoung1981:


src/SFA.DAS.EmployerAccounts/Data/EmployerAccountRepository.cs
```diff
@@ -82,56 +82,48 @@ public async Task<Account> GetAccountByHashedId(string hashedAccountId)
...
-                    ApprenticeshipEmployerType = (ApprenticeshipEmployerType) ac.ApprenticeshipEmployerType
-                }).FirstOrDefaultAsync();
-
-            if (accountDetail == null)
-            {             
+        {
+            var account = await _db.Value.Accounts.Include(x => x.AccountLegalEntities.Select(y => y.Agreements)).SingleAsync(x => x.HashedId == hashedAccountId);
+
+            if (account == null)
+            {
```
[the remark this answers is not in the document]
MichaelYoung1981: You are of course correct...I spent a bit of time thinking about whether an invalid hashed account id should return null or throw and exception and came to the conclusion that it is indeed exceptional behaviour but then I decided I didn't want to change the behaviour of the class so left the null check in but forgot to change back to SingleOrDefaultAsync!

src/SFA.DAS.EmployerAccounts/Data/EmployerAccountRepository.cs
```diff
@@ -82,56 +82,48 @@ public async Task<Account> GetAccountByHashedId(string hashedAccountId)
...
+
+            var activeLegalEntities = account.AccountLegalEntities.Where(x =>
+                x.Deleted == null && x.Agreements.Any(ea =>
+                    ea.StatusId == EmployerAgreementStatus.Pending ||
+                    ea.StatusId == EmployerAgreementStatus.Signed));
+
+            accountDetail.LegalEntities = activeLegalEntities.Select(x => x.Id).ToList();
+            accountDetail.AccountAgreementTypes = account.AccountLegalEntities.SelectMany(x => x.Agreements).Select(x => x.Template.AgreementType).Distinct().ToList();
+
+            var ownerEmailTask = _db.Value.Memberships
```
CodeWolf86: All looks good to me in terms of the query as it mimics what was there before just optimised. Only comment is that given that both Account History, and Memberships we are querying by accountId, is it worth modifying the entity models to reflect these relationships? which would mean you could include account history and membership in your top query, do one join up front and then you will have all the information you need?

MichaelYoung1981: I did consider it but bottled making the change!  Believe it or not, these changes were meant to be as light touch as possible. 😂

satya-sanikommu:
This is not longer required, as it will be merged as part of #2013 . Closing this PR

PR #1929, opened 2019-11-07
css and html for support user banner

DH02:
Generated files from _header.scss additions and html partial for the new banner.


DH02:
Not required.

PR #1944, opened 2019-11-15
CON-1015-add-support-user-auditing

Stephek:


src/SFA.DAS.EmployerAccounts/Services/AuditApiClientForSupportUser.cs
```diff
@@ -0,0 +1,43 @@
...
+        public AuditApiClientForSupportUser(IAuditApiClient client, IAuthenticationService authenticationService)
+        {
+            _client = client;
+            _authenticationService = authenticationService;
+        }
+        public Task Audit(AuditMessage message)
+        {
+            if (_authenticationService.IsSupportUser())
+            {
+                var impersonatedUser = _authenticationService.GetClaimValue(DasClaimTypes.Id);
```
cofaulco: I don't think that these two claims are being populated with the impersonated users information as they would normally be when coming through the staff support user route.

src/SFA.DAS.EmployerAccounts/Commands/AuditCommand/CreateAuditCommandValidator.cs
```diff
@@ -24,7 +24,10 @@ public ValidationResult Validate(CreateAuditCommand item)
 
             if (item.EasAuditMessage.ChangedProperties==null || !item.EasAuditMessage.ChangedProperties.Any())
             {
-                validationResult.AddError(nameof(item.EasAuditMessage.ChangedProperties));
+                if (item.EasAuditMessage.Category == null || !item.EasAuditMessage.Category.Equals("VIEW", StringComparison.InvariantCultureIgnoreCase))
```
cofaulco: thought we had gotten rid of the requirement to audit when the support user view something?

src/SFA.DAS.EmployerAccounts/Services/AuditApiClientForSupportUser.cs
Stephek: These were the claims that were setup on the identity when Vas logged on as a staff user.

src/SFA.DAS.EmployerAccounts/Commands/AuditCommand/CreateAuditCommandValidator.cs
Stephek: No it's acceptance criteria 1 in the story.

cofaulco: ok

src/SFA.DAS.EmployerAccounts/Services/AuditApiClientForSupportUser.cs
cofaulco: ok I think we need to look at this when testing alongside the other branches because I'm not sure these claims should be being set when logging in via staff idams

cofaulco: approving as is, but may need reviewing during testing

PR #1954, opened 2019-11-21
Logging Upgrade to Finance Proj caused binding exception in finance api

CodeWolf86:


src/SFA.DAS.EmployerFinance.Api/Web.config
```diff
@@ -1,290 +1,358 @@
...
+			<dependentAssembly>
+				<assemblyIdentity name="Microsoft.Extensions.Primitives" publicKeyToken="ADB9793829DDAE60" culture="neutral"/>
+				<bindingRedirect oldVersion="0.0.0.0-2.2.0.0" newVersion="2.2.0.0"/>
+			</dependentAssembly>
+			<dependentAssembly>
+				<assemblyIdentity name="Microsoft.Extensions.Options" publicKeyToken="ADB9793829DDAE60" culture="neutral"/>
+				<bindingRedirect oldVersion="0.0.0.0-2.2.0.0" newVersion="2.2.0.0"/>
+			</dependentAssembly>
+			<dependentAssembly>
+				<assemblyIdentity name="Microsoft.Extensions.Logging" publicKeyToken="ADB9793829DDAE60" culture="neutral"/>
```
CodeWolf86: Looks like a lot of change here, and it is, but visual studio had a warning offering to fix binding redirects so chose to do that rather than just fix the one we were seeing which was the logging one above.

PR #1956, opened 2019-11-21
CON-1042-preserve-ga-parameter-in-redirect

Stephek:


src/SFA.DAS.EAS.Web/Controllers/HomeController.cs
```diff
@@ -8,13 +8,15 @@ namespace SFA.DAS.EAS.Web.Controllers
...
+        private const string GoogleTag = "_ga";
+
         [Route("~/")]
         [Route]
         [Route("Index")]
         public ActionResult Index()
         {
-            return Redirect(Url.EmployerAccountsAction("service/index", false));
-        }
+            return Redirect(Url.EmployerAccountsAction($"service/index?{GetTrackerQueryString()}", false));
```
dev4valtech: I think the requirement was to persist it across all the pages.

Stephek: This is the landing page and the story only specifies that this is where the tracking id is preserved.

cofaulco: I can't find an example where we are manually persisting this across all pages via url or cookie across other applications, so going to approve this as is and we can pick up with GA guys once in.

PR #1957, opened 2019-11-22
Aml 3828 grant provider permissions

Howesy77:


src/SFA.DAS.EmployerAccounts.UnitTests/Commands/SignEmployerAgreementTests/WhenISignAnEmployerAgreement.cs
```diff
@@ -95,6 +97,8 @@ public void Setup()
             _genericEventFactory = new Mock<IGenericEventFactory>();
             _mediator = new Mock<IMediator>();
 
+            _mediator.Setup(x => x.SendAsync(It.IsAny<GetUserByRefQuery>())).ReturnsAsync(new GetUserByRefResponse { User = new User { CorrelationId = "CORRELATION_ID" } });
```
CodeWolf86: Any way we can tighten the `GetUserByRefQuery` part of the setup?

CodeWolf86: Sorted it

PR #1958, opened 2019-11-22
CON-1089-make-duplicate-reservation-event-handling-idempotent

Stephek:


Stephek:
> should we log a warning?

Yes. Now added.

PR #1962, opened 2019-11-25
CON-943 Removed levy tasks for non-levy employer

dev4valtech:


src/SFA.DAS.EmployerAccounts.Web/Orchestrators/EmployerTeamOrchestrator.cs
```diff
@@ -227,6 +227,12 @@ public async Task<OrchestratorResponse<AccountDashboardViewModel>> GetAccount(st
                     ShowSavedFavourites = _authorizationService.IsAuthorized("EmployerFeature.HomePage"),
                     ReservationsCount = reservationsResponse.Reservations.Count()
                 };
+                
+                if (viewModel.ApprenticeshipEmployerType == ApprenticeshipEmployerType.NonLevy)
+                {
+                    var levyReminderTask = tasks.FirstOrDefault(t => t.Type == "LevyDeclarationDue");
+                    tasks.Remove(levyReminderTask);
+                }
```
MichaelYoung1981: Wouldn't it be better if the Tasks worker didn't create the task for NonLevy employers rather than filtering it out after retrieval?

dev4valtech: From my understanding, this a reminder task with single instance. So there are no tasks getting created, it just checks if the user has already dismissed one for the month or not. So there is nothing I can use to circumnavigate that. Please correct me if I am wrong.

MichaelYoung1981:
You could add the ApprenticeshipEmployerType to the MonthlyReminderTasks in
das-tasks and filter it in the stored proc (GetMonthlyReminderTasks).  It'd
mean passing the ApprenticeshipEmployerType from EmployerAccounts through
to the API but it feels like it'd be in the right place then.

Not a tiny change granted. 😀

src/SFA.DAS.EmployerAccounts.Web/Orchestrators/EmployerTeamOrchestrator.cs
MichaelYoung1981: You could add the ApprenticeshipEmployerType to the MonthlyReminderTasks in das-tasks and filter it in the stored proc (GetMonthlyReminderTasks).  It'd mean passing the ApprenticeshipEmployerType from EmployerAccounts through to the API but it feels like it'd be in the right place then.

Not a tiny change granted. 😀

dev4valtech: Yes could do that. By naming, the Tasks seemed more generic, adding this will make it more specific, is that alright?

PR #1964, opened 2019-11-27
Aml 4038 in account sign agreement link

Howesy77:


src/SFA.DAS.EmployerAccounts.Web/Controllers/EmployerAgreementController.cs
```diff
@@ -128,6 +129,13 @@ public async Task<ActionResult> SignAgreement(GetEmployerAgreementRequest reques
             var response = await _mediator.SendAsync(request);
             var viewModel = _mapper.Map<GetEmployerAgreementResponse, EmployerAgreementViewModel>(response);
 
+            var legalentities = await _mediator.SendAsync(new GetAccountLegalEntitiesByHashedAccountIdRequest
+            {
+                HashedAccountId = request.HashedAccountId
+            });
+
+            viewModel.LegalEntitiesCount = legalentities.LegalEntities.Count;
```
MichaelYoung1981: With my performance hat on...this is a very inefficient way of getting the legal entity count.  Could you write a specific query to return just the count?

Howesy77: Yes can do

Howesy77: How about using the AccountStats query?

MichaelYoung1981: Does too much unfortunately.

Howesy77: Ok changes made

PR #1977, opened 2020-01-06
Removed underline from ESFA text

Howesy77:


src/SFA.DAS.EmployerAccounts.Web/Views/Shared/_CookieBanner.cshtml
```diff
@@ -3,7 +3,7 @@
     <div class="das-cookie-banner">
         <div class="das-cookie-banner__wrapper govuk-width-container">
             <h2 class="heading-large das-cookie-banner__header">Cookies</h2>
-            <p class="das-cookie-banner__message">The <abbr title="Education and Skills Funding Agency">ESFA</abbr> uses cookies to make the site simpler.</p>
+            <p class="das-cookie-banner__message">The <span title="Education and Skills Funding Agency">ESFA</span> uses cookies to make the site simpler.</p>
```
CodeWolf86: Rather than remove the `abbr` tag which is correct for abbreviations can we not either remove the underline from abbreviation CSS, or add a css rule for `.das-cookie-banner__message abbr` where `border-bottom:none` is set instead? 

image

Howesy77: I suggest we push this to a FE dev then as the CDN will need updating?

PR #1981, opened 2020-01-08
CON-1190-split-privacy-and-cookies-link

Stephek:


src/SFA.DAS.EmployerFinance.Web/Extensions/UrlHelperExtensions.cs
```diff
@@ -14,6 +14,19 @@ public static string EmployerAccountsAction(this UrlHelper helper, string path,
             return includedAccountId ? AccountAction(helper, baseUrl, path) : Action(baseUrl, path);
         }
 
+        public static string AccountsAction(this UrlHelper helper, string controller, string action, bool includedAccountId = true)
```
cofaulco: Name of this method should be consistent with the one above as it seems to be simply an overload for EmployerAccountsAction()

src/SFA.DAS.EmployerFinance.Web/Extensions/UrlHelperExtensions.cs
```diff
@@ -14,6 +14,19 @@ public static string EmployerAccountsAction(this UrlHelper helper, string path,
...
+        {
+            var configuration = DependencyResolver.Current.GetService<EmployerFinanceConfiguration>();
+            var baseUrl = configuration.EmployerAccountsBaseUrl;
+            if (includedAccountId)
+            {
+                var hashedAccountId = helper.RequestContext.RouteData.Values[ControllerConstants.AccountHashedIdRouteKeyName];
+                return Action(baseUrl, controller, action, hashedAccountId?.ToString());
```
cofaulco: does the Action() method handle hashedAccountId being null or empty?

PR #1982, opened 2020-01-08
CON-1167 : Finance - Cookie banner

VasanthaKasirajan3008:


src/SFA.DAS.EmployerFinance.Web/Extensions/UrlHelperExtensions.cs
```diff
@@ -14,6 +14,19 @@ public static string EmployerAccountsAction(this UrlHelper helper, string path,
             return includedAccountId ? AccountAction(helper, baseUrl, path) : Action(baseUrl, path);
         }
 
+        public static string AccountsAction(this UrlHelper helper, string controller, string action, bool includedAccountId = true)
```
cofaulco: Made the same comment on Kevins PR, should this be EmployerAccountsAction() and simply be an overload?

src/SFA.DAS.EmployerFinance.Web/Controllers/HomeController.cs
```diff
@@ -37,6 +38,14 @@ public ActionResult Privacy()
             return Redirect(Url.LegacyEasAction("service/privacy"));
         }
 
+        [HttpGet]
+        [Route("{HashedAccountId}/cookieConsent", Order = 0)]
+        [Route("cookieConsent", Order = 1)]
+        public ActionResult CookieConsent()
```
cofaulco: Did we need to create a controller action? Other places we just point the link directly to the relevant url in EmployerAccounts

VasanthaKasirajan3008: Its to keep the consistency with the they way it works for privacy.

PR #1985, opened 2020-01-09
Aml 4077

Howesy77:


src/SFA.DAS.EAS.Api/Controllers/UserController.cs
```diff
@@ -0,0 +1,32 @@
...
+    public class UserController : ApiController
+    {
+        private readonly IEmployerAccountsApiService _apiService;
+
+        public UserController(IEmployerAccountsApiService apiService)
+        {
+            _apiService = apiService;
+        }
+
+        [Route("")]
```
CodeWolf86: I know this is legit, but it does look a bit hacky, can we not get away with just `[HttpGet]` and then allow it to inherit the route from the controller attribute. I know i'm being picky.....

src/SFA.DAS.EmployerAccounts.UnitTests/Queries/GetUserByEmailTests/WhenIGetAUser.cs
```diff
@@ -0,0 +1,71 @@
...
+            _repository = new Mock<IUserAccountRepository>();
+
+            _repository.Setup(x => x.Get(It.IsAny<string>())).ReturnsAsync(_user);
+
+            Query = new GetUserByEmailQuery { Email = "test@test.com" };
+            RequestHandler = new GetUserByEmailQueryHandler(_repository.Object, RequestValidator.Object);
+        }
+
+        [Test]
+        public async Task ThenIShouldThrowExceptionIfTheUserCannotBeFound()
```
CodeWolf86: Is the test title right? as it returns null rather than throws? or is the test meant to be for the validator where an email isn't passed and so throws InvalidRequest?

src/SFA.DAS.EmployerAccounts/Queries/GetUserByEmail/GetUserByEmailQueryValidator.cs
```diff
@@ -0,0 +1,26 @@
...
+            {
+                result.AddError(nameof(item.Email), "User email has not been supplied");
+            }
+
+            return result;
+        }
+
+        public Task<ValidationResult> ValidateAsync(GetUserByEmailQuery item)
+        {
+            throw new NotImplementedException();
```
CodeWolf86: can we call the above method rather than just not implemented?

PR #2001, opened 2020-01-22
CON-1144_Promote Tier1 user

shomavg:


src/SFA.DAS.EmployerAccounts.Web/Web.config
```diff
@@ -19,7 +19,8 @@
     <add key="TokenCertificateThumbprint" value="6B7ACC520305BFDB4F7252DAEB2177CC091FAAE1"/>
     <add key="UnobtrusiveJavaScriptEnabled" value="false"/>
     <add key="webpages:Enabled" value="false"/>
-    <add key="webpages:Version" value="3.0.0.0"/>    
+    <add key="webpages:Version" value="3.0.0.0"/>
+    <add key="SupportConsoleUser" value="Tier1User,Tier2User"/>
```
cofaulco: This config value needs to be read from table storage in the EmployerAccountConfiguration

src/SFA.DAS.EmployerAccounts.Web/Helpers/AccountTaskHelper.cs
```diff
@@ -20,5 +27,22 @@ public static int GetTaskPriority(AccountTask task)
                 default: return int.MaxValue; //if its an usupported type we place it last
             }
         }
+
+        public static bool IsSupportConsoleUser(IPrincipal user)
```
cofaulco: This doesn't belong in this class. This is specific to account tasks

src/SFA.DAS.EmployerAccounts.Web/Helpers/AccountTaskHelper.cs
```diff
@@ -20,5 +27,22 @@ public static int GetTaskPriority(AccountTask task)
...
+        {
+            var configuration = DependencyResolver.Current.GetService<EmployerAccountsConfiguration>();
```
cofaulco: This should be some sort of re-usable service that we can inject and test

src/SFA.DAS.EmployerAccounts.Web/Startup.cs
```diff
@@ -214,6 +214,9 @@ private Task OnSecurityTokenValidated(SecurityTokenValidatedNotification<WsFeder
                 }
                 else if (notification.AuthenticationTicket.Identity.HasClaim(serviceClaimType, "ESS"))
                 {
+                    Logger.Debug("Adding Tier1 Role");
```
cofaulco: I don't think we need to add this additional claim. Tier2 users will end up with 3 Role claims one of which being Tier1User. Let's just use Tier2User and ConsoleUser for now.

src/SFA.DAS.EmployerAccounts/Extensions/AuthenticationServiceExtensions.cs
```diff
@@ -1,13 +1,20 @@
...
 
 namespace SFA.DAS.EmployerAccounts.Extensions
 {
     public static class AuthenticationServiceExtensions
     {
         public static bool IsSupportUser(this IAuthenticationService authenticationService)
-        {   
-            return authenticationService.HasClaim(ClaimsIdentity.DefaultRoleClaimType, Constants.Tier2User);
+        {
+            var configuration = DependencyResolver.Current.GetService<EmployerAccountsConfiguration>();
```
cofaulco: I think this is the neatest of the several options in this PR.

src/SFA.DAS.EmployerAccounts/Services/AuditApiClientForSupportUser.cs
```diff
@@ -6,24 +6,24 @@
...
 using System.Collections.Generic;
-using SFA.DAS.EmployerAccounts.Configuration;
 
 namespace SFA.DAS.EmployerAccounts.Services
 {
     public class AuditApiClientForSupportUser : IAuditApiClient
     {
         private readonly IAuditApiClient _client;
+        private readonly IUserContext _userContext;
         private readonly IAuthenticationService _authenticationService;
```
Stephek: The IAuthenticationService dependency could be removed from this class if either the Email and UserId are moved to the IUserContext interface or the GetClaimValue method is exposed by the IUserContext interface.

PR #2002, opened 2020-01-24
Remove code to trigger expiry of funds processing until fix has been …

cofaulco:
…implemented for the calculation


cofaulco:
Merging hotfix post deployment to ensure job is not accidentally switched back on by deployment of master to PROD

PR #2014, opened 2020-02-13
CON-1358 Feature add an apprentice

ChrisJWoodcock:


src/SFA.DAS.EmployerAccounts.Web/ViewModels/Row1Panel1ViewModel.cs
```diff
@@ -0,0 +1,21 @@
+﻿using SFA.DAS.EmployerAccounts.Web.Extensions;
+using System;
+
+namespace SFA.DAS.EmployerAccounts.Web.ViewModels
+{
+    public class Row1Panel1ViewModel
+    {
```
cofaulco: Does this need to be merged with the call to action view model now?

src/SFA.DAS.EmployerAccounts/DependencyResolution/EncodingRegistry.cs
```diff
@@ -0,0 +1,13 @@
+﻿using SFA.DAS.Encoding;
+using StructureMap;
+
+namespace SFA.DAS.EmployerAccounts.DependencyResolution
+{
+    public class EncodingRegistry : Registry
+    {
+        public EncodingRegistry()
+        {
+            For<IEncodingService>().Use<EncodingService>().Singleton();
```
cofaulco: We may not want to mix usage of the EncodingService and HashingService

src/SFA.DAS.EmployerAccounts.Web/ViewModels/Row1Panel1ViewModel.cs
ChrisJWoodcock: Yes - but that is in the 1261 branch currently

src/SFA.DAS.EmployerAccounts/DependencyResolution/EncodingRegistry.cs
ChrisJWoodcock: This one is for you@VasanthaKasirajan3008

VasanthaKasirajan3008: We need to use Encoding Service to consume Commitments Api.

cofaulco:
Draft PR other related PR's all now merged

PR #2018, opened 2020-02-20
Con 1384 use injected config

cofaulco:


src/SFA.DAS.EAS.Web/Controllers/EmployerCommitmentsController.cs
```diff
@@ -10,6 +11,13 @@ namespace SFA.DAS.EAS.Web.Controllers
     [RoutePrefix("accounts/{hashedaccountId}/apprentices")]
     public class EmployerCommitmentsController : Controller
     {
+        private readonly EmployerApprenticeshipsServiceConfiguration _easConfig;
+
+        public EmployerCommitmentsController(EmployerApprenticeshipsServiceConfiguration easConfig) : base()
+        {
+            _easConfig = easConfig;
```
VasanthaKasirajan3008: will it be good to rename to  employerApprenticeshipsServiceConfiguration ?

PR #2030, opened 2020-03-02
[CON-1399] For some reason, payments seem to be being returned from t…

cofaulco:
…he PaymentsEventsApi with a StandardCode of 0 rather than NULL when they in fact relate to a Framework. The code having previously always expected a NULL value for this property was checking HasValue(). Changed to check for a zero value if NULL and to assume that it is a framework otherwise. Also changed to log if both frameworkCode and standardCode not set


cofaulco:
> 👍 are we doing anything to correct the existing data?

Yes we have a separate ticket for a data fix CON-1544

PR #2032, opened 2020-03-06
CON-1495 MA banner

shomavg:


src/SFA.DAS.EmployerAccounts.Web/Helpers/HtmlHelperExtensions.cs
```diff
@@ -30,19 +33,49 @@ public static bool IsAuthorized(this HtmlHelper htmlHelper, string featureType)
...
+                var legalEntityAgreements = employerAgreements.GroupBy(ea => ea.LegalEntity.AccountLegalEntityId);
+
+                foreach (var legalEntityAgreement in legalEntityAgreements)
+                {
+                    var latestSignedAgreement = legalEntityAgreement
+                        .Where(lea => lea.HasSignedAgreement)
+                        .OrderByDescending(lea => lea.Signed.VersionNumber)
+                        .FirstOrDefault();
+
+                    if (latestSignedAgreement?.Signed.VersionNumber != 3) return true;
```
cofaulco: this should return false i.e hasn't signed v3 agreement.

src/SFA.DAS.EmployerAccounts.Web/Helpers/HtmlHelperExtensions.cs
```diff
@@ -30,19 +33,49 @@ public static bool IsAuthorized(this HtmlHelper htmlHelper, string featureType)
...
+                }
+            }
+            return false;
```
cofaulco: should return true here because all legal entities have a signed v3 agreement

src/SFA.DAS.EmployerAccounts.Web/Views/Shared/_Layout.cshtml
```diff
@@ -228,11 +228,20 @@
...
     }
-            <main role="main" id="content">
-                @RenderSection("breadcrumb", required: false)
-                @Html.Partial(@"_SuccessMessage", Model as OrchestratorResponse)
-                @RenderBody()
-            </main>
+    <main role="main" id="content">
+        @RenderSection("breadcrumb", required: false)
+        @{
+            var userId = Html.ViewContext.RequestContext.HttpContext.GetOwinContext().Authentication.User.Claims.
```
cofaulco: Might be best to keep all this inside the htmlHelper extension as we will be less likely to miss it when removing this code.

PR #2060, opened 2020-04-07
Con 1492 organisations & agreements

Howesy77:


src/SFA.DAS.EmployerAccounts.Web/Controllers/EmployerAgreementController.cs
```diff
@@ -327,5 +310,12 @@ public async Task<ActionResult> WhenDoYouWantToView(int? choice, string agreemen
                 }
             }
         }
+
+        [HttpGet]
+        [Route("agreements/{accountLegalEntityHashedId}/viewallagreements")]
+        public ActionResult ViewAllAgreements(string hashedAccountId, string accountLegalEntityHashedId)
+        {
+           throw new NotImplementedException();
```
ChrisJWoodcock, line 318: In this a placeholder for Vas's story?

ChrisJWoodcock: If you can call this it shouldn't throw it should put up a message - so this story could be shipped independently

Howesy77: Ok has been reverted to previous functionality

src/SFA.DAS.EmployerAccounts.Web/Controllers/EmployerAgreementController.cs
```diff
@@ -240,63 +240,46 @@ public async Task<ActionResult> GetSignedPdfAgreement(string agreementId, string
...
+            var model = await _orchestrator.GetConfirmRemoveOrganisationViewModel(accountLegalEntityHashedId, hashedAccountId, OwinWrapper.GetClaimValue(ControllerConstants.UserRefClaimKeyName));
+
+            return View(model.Data.CanBeRemoved ? ControllerConstants.ConfirmRemoveOrganisationActionName : ControllerConstants.CannotRemoveOrganisationViewName, model.Data);
         }
 
-        [HttpGet]
-        [Route("agreements/remove/{agreementId}")]
-        public async Task<ActionResult> ConfirmRemoveOrganisation(string agreementId, string hashedAccountId)
+        [HttpPost]
+        [Route("agreements/{accountLegalEntityHashedId}/remove")]
```
Stephek, line 252: Shouldn't the route be?

accounts/{HashedAccountId}/organisations/{accountLegalEntityHashedId}/agreements/remove ?

Howesy77: No, potentially it should be in a different controller now as the remove is done by account legal entity and not agreement.

PR #2063, opened 2020-04-14
[CON-1719] Should handle add apprentice api call failures gracefully.…

cofaulco:
… This is replicating the logic used for the reservations and vacancies calls.

### DIFF
--- a/src/SFA.DAS.EmployerAccounts/Queries/GetApprenticeship/GetApprenticeshipsResponse.cs
+++ b/src/SFA.DAS.EmployerAccounts/Queries/GetApprenticeship/GetApprenticeshipsResponse.cs
@@ -6,5 +6,6 @@ namespace SFA.DAS.EmployerAccounts.Queries.GetApprenticeship
     public class GetApprenticeshipsResponse
     {
         public IEnumerable<Apprenticeship> Apprenticeships { get; set; }
+        public bool HasFailed { get; set; }
     }
 }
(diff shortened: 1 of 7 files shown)


src/SFA.DAS.EmployerAccounts/Queries/GetSingleCohort/GetSingleCohortResponse.cs
```diff
@@ -5,5 +5,6 @@ namespace SFA.DAS.EmployerAccounts.Queries.GetSingleCohort
     public class GetSingleCohortResponse
     {
         public Cohort Cohort { get; set; }
+        public bool HasFailed { get; set; }
```
Stephek, line 8: If this is to be common across all response types then it should be put into a base class.

src/SFA.DAS.EmployerAccounts/Queries/GetApprenticeship/GetApprenticeshipsResponse.cs
```diff
@@ -6,5 +6,6 @@ namespace SFA.DAS.EmployerAccounts.Queries.GetApprenticeship
     public class GetApprenticeshipsResponse
     {
         public IEnumerable<Apprenticeship> Apprenticeships { get; set; }
+        public bool HasFailed { get; set; }
```
Stephek, line 9: If this is to be common across all response types then it should be put into a base class.

PR #2066, opened 2020-04-17
CON-1701 raise event on levy change

Howesy77:


src/SFA.DAS.EmployerFinance.Messages/Events/LevyAddedToAccountEvent.cs
```diff
@@ -2,7 +2,7 @@
 
 namespace SFA.DAS.EmployerFinance.Messages.Events
 {
-    public class LevyAddedToAccount : Event
+    public class LevyAddedToAccountEvent : Event
```
cofaulco, line 5: Due to how subscriptions currently work I think we should leave the name of this event, as consumers expect the old name and namespace. I think all consumers would have to update in order to consume these events

src/SFA.DAS.EmployerAccounts.MessageHandlers/EventHandlers/EmployerFinance/RefreshEmployerLevyDataCompletedEventHandler.cs
```diff
@@ -9,23 +12,30 @@ namespace SFA.DAS.EmployerAccounts.MessageHandlers.EventHandlers.EmployerFinance
...
-                    message.Created,
-                    string.Empty,
-                    string.Empty));
+            await _messagePublisher.PublishAsync(new RefreshEmployerLevyDataCompletedMessage(
+                message.AccountId,
+                message.LevyImported,
+                message.PeriodMonth,
+                message.PeriodYear,
+                message.Created,
+                string.Empty,
```
cofaulco, line 31: Where are we still publishing this obsolete message to?

PR #2067, opened 2020-04-22
CON-1749-check version of pending agreement

shomavg:


src/SFA.DAS.EmployerAccounts/Authorisation/EmployerFeatureAuthorisationHandler.cs
```diff
@@ -51,7 +52,9 @@ public async Task<AuthorizationResult> GetAuthorizationResult(IReadOnlyCollectio
                         .DefaultIfEmpty()
                         .Min();
 
-                    if (minAgreementVersion < featureToggle.EnabledByAgreementVersion || agreements.EmployerAgreements.Any(a => a.StatusId == Models.EmployerAgreement.EmployerAgreementStatus.Pending))
+                    if (minAgreementVersion < featureToggleEnabledByAgreementVersion || 
+                        agreements.EmployerAgreements.Any(a => a.StatusId == Models.EmployerAgreement.EmployerAgreementStatus.Pending &&
+                                                               a.TemplateId == featureToggleEnabledByAgreementVersion))
```
Howesy77, line 57: The template Id isn't the same as the version but I don't think this would be the desired fix anyway.

PR #2069, opened 2020-04-23
CON-1721 log exception

shomavg:


src/SFA.DAS.EmployerAccounts.ReadStore/SFA.DAS.EmployerAccounts.ReadStore/Application/Commands/CreateAccountUserCommandHandler.cs
```diff
@@ -10,27 +12,35 @@ namespace SFA.DAS.EmployerAccounts.ReadStore.Application.Commands
...
+                    user.Recreate(request.Role, request.Created, request.MessageId);
+                    await _accountUsersRepository.Update(user, null, cancellationToken);
+                }
             }
-            else
+            catch (UserNotRemovedException exception)
             {
-                user.Recreate(request.Role, request.Created, request.MessageId);
-                await _accountUsersRepository.Update(user, null, cancellationToken);
+                _logger.LogWarning("User has not been removed", exception);
```
Stephek, line 42: Shouldn't the exception be re-thrown rather than swallowed?

src/SFA.DAS.EmployerAccounts.ReadStore/SFA.DAS.EmployerAccounts.ReadStore/Application/Commands/CreateAccountUserCommandHandler.cs
```diff
@@ -10,27 +12,35 @@ namespace SFA.DAS.EmployerAccounts.ReadStore.Application.Commands
...
+                    await _accountUsersRepository.Add(user, null, cancellationToken);
+                }
+                else
+                {
+                    user.Recreate(request.Role, request.Created, request.MessageId);
+                    await _accountUsersRepository.Update(user, null, cancellationToken);
+                }
             }
-            else
+            catch (UserNotRemovedException exception)
```
Stephek, line 40: Just for info what throws this exception?

cofaulco, line 40: We are catching this custom Exception but never throwing it

src/SFA.DAS.EmployerAccounts.ReadStore/SFA.DAS.EmployerAccounts.ReadStore/Models/AccountUser.cs
```diff
@@ -141,7 +142,7 @@ private void EnsureUserHasBeenRemoved()
         {
             if (Removed == null)
             {
-                throw new InvalidOperationException("User has not been removed");
+                throw new UserNotRemovedException("User has not been removed");
```
Stephek, line 145: I think the issue raised with the ticket is because of the EnsureUserHasBeenRemoved method.

If my understanding is correct...

The process is creating a user that either does not exist or has been removed?

If the user already exists then we try to re-create the user but only if the recreate date is greater than the created date.

The first check is to check that the user has actually been previously removed.  
Looking at the user.Remove method if a user has been removed then the removed property is set to the current date.

Therefore when we are checking that the user to be recreated has actually been removed then we would expect the Removed property to have a value rather than null.

So the EnsureUserHasBeenRemoved is incorrect.  It should be 

if (Removed.HasValue)
{
       throw new UserNotRemovedException("User has not been removed");
}

That is the only change required in this ticket.  

Please correct me if my understanding is wrong.

src/SFA.DAS.EmployerAccounts.ReadStore/SFA.DAS.EmployerAccounts.ReadStore/Application/Commands/CreateAccountUserCommandHandler.cs
```diff
@@ -10,28 +12,38 @@ namespace SFA.DAS.EmployerAccounts.ReadStore.Application.Commands
...
-            var user = await _accountUsersRepository.CreateQuery().SingleOrDefaultAsync(x => x.UserRef == request.UserRef && x.AccountId == request.AccountId, cancellationToken);
-
-            if (user == null)
+            try
             {
-                user = new AccountUser(request.UserRef, request.AccountId, request.Role, request.Created, request.MessageId);
-                await _accountUsersRepository.Add(user, null, cancellationToken);
+                var user = await _accountUsersRepository.CreateQuery().SingleOrDefaultAsync(x => x.UserRef == request.UserRef && x.AccountId == request.AccountId, cancellationToken);
+
+                if (user == null)
```
Stephek, line 29: We check if the user exists, and if it doesn't then assume that he user should be recreated.  what we don't check is if the user has already been recreated.

So the logic should be 

if(!UserExists)
{
   Create User
}
else if (UserAlreadyRecreated)
{
 Update Roles
}
else
{
 RecreateUser
}

Then the exception does not need to be caught.

cofaulco: Ok I think I agree with the above

cofaulco: so if (user == null) {
    CreateUser()
} else if (user.Deleted == null) {
   UpdateRoles()
} else if (user.Delete.HasValue) {
   RecreateUser();
}

PR #2070, opened 2020-04-27
CON-1746-content-change

Stephek:


src/SFA.DAS.EmployerAccounts.Web/Controllers/EmployerTeamController.cs
```diff
@@ -58,6 +58,12 @@ public async Task<ActionResult> Index(string hashedAccountId, string reservation
                 ViewBag.HideNav = true;
             }
 
+            var flashMessage = GetFlashMessageViewModelFromCookie();
+            if (flashMessage != null)
+            {
+                response.FlashMessage = flashMessage;
+            }
+
```
cofaulco, line 66: this is being done in GetAccountInformation() called above

Stephek: Ah yes.  Will remove.

PR #2074, opened 2020-04-29
CON-1764 content banner

shomavg:


src/SFA.DAS.EmployerAccounts.Web/Extensions/HtmlHelperExtensions.cs
```diff
@@ -132,5 +133,17 @@ public static IFooterViewModel GetFooterViewModel(this HtmlHelper html, bool use
             useLegacyStyles: useLegacyStyles
             );
         }
+
+        public static MvcHtmlString GetClientContentByType(this HtmlHelper html, string type, string clientId)
```
Stephek, line 137: Do we need to pass in the client Id since we could resolved it from the config?

src/SFA.DAS.EmployerAccounts/Interfaces/ICacheStorageService.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System.Threading.Tasks;
+
+namespace SFA.DAS.EmployerAccounts.Interfaces
+{
+    public interface ICacheStorageService
+    {
+        Task<T> RetrieveFromCache<T>(string key);
```
Stephek, line 7: As this interface is to do with CacheStorage the method names don't need to specify that data is being retrieved form the cache.  Maybe just call the methods Get, Save and Delete to remove the tautology.

src/SFA.DAS.EmployerAccounts/Interfaces/IClientContentApiClient.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System.Threading.Tasks;
+using SFA.DAS.EmployerAccounts.Queries.GetClientContent;
+
+namespace SFA.DAS.EmployerAccounts.Interfaces
+{
+    public interface IClientContentApiClient
+    {
+        Task<string> GetContentByClientId(ContentType type, string clientId);
```
Stephek, line 8: Could the method just be Get ?  since that is what it is doing.  It's explicit that it's by ClientId since that is a required parameter.

src/SFA.DAS.EmployerAccounts/Queries/GetClientContent/GetClientContentRequestHandler.cs
```diff
@@ -0,0 +1,103 @@
...
+                throw new InvalidRequestException(validationResult.ValidationDictionary);
+            }
+
+            try
+            {
+                var type = (ContentType) Enum.Parse(typeof(ContentType), message.ContentType, true);
+
+                var cacheKey = message.ClientId;
+
+                if (!string.IsNullOrWhiteSpace(cacheKey))
```
Stephek, line 43: This check has already been done in the validator.  If the GetClientContentRequest  passes validation then we can assume it has a ClientId.  Please remove this check.

src/SFA.DAS.EmployerAccounts/Queries/GetClientContent/GetClientContentRequestHandler.cs
```diff
@@ -0,0 +1,103 @@
...
+
+                            ContentBanner = result
+                        };
+                    }
+                }
+                var cachedContentBanner = await _service.GetContentByClientId(type, message.ClientId);
+
+                if (cachedContentBanner == null)
+                {
+                    throw new CachedContentBannerNotFoundException(message.ClientId);
```
Stephek, line 61: Shouldn't the service throw a ContentNotFound exception or return a status that can be checked?

This error is misleading since the error doesn't happen because the cached item can't be found.

src/SFA.DAS.EmployerAccounts/Queries/GetClientContent/GetClientContentRequestHandler.cs
```diff
@@ -0,0 +1,103 @@
...
+            try
+            {
+                var type = (ContentType) Enum.Parse(typeof(ContentType), message.ContentType, true);
+
+                var cacheKey = message.ClientId;
+
+                if (!string.IsNullOrWhiteSpace(cacheKey))
+                {
+                    var result =
+                        await _cacheStorageService.RetrieveFromCache<string>(cacheKey);
```
Stephek, line 46: It would be cleaner code to have a TryGet method on the cacheStorageService.

i.e. bool TryGet<T>(string key, out T result)

src/SFA.DAS.EmployerAccounts/Queries/GetClientContent/GetClientContentResponse.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace SFA.DAS.EmployerAccounts.Queries.GetClientContent
+{
+    public class GetClientContentResponse
+    {
+        public string ContentBanner { get; set; }
```
Stephek, line 5: This is just Content, not necessarily a banner.

src/SFA.DAS.EmployerAccounts/Services/CacheStorageService.cs
```diff
@@ -0,0 +1,39 @@
...
+
+        public CacheStorageService(IDistributedCache distributedCache)
+        {
+            _distributedCache = distributedCache;
+        }
+
+        public async Task SaveToCache<T>(string key, T item, int expirationInHours)
+        {
+            var json = JsonConvert.SerializeObject(item);
+            await _distributedCache.SetCustomValueAsync(key, json, new TimeSpan(1, 0, 0));
```
Stephek, line 21: The cache period timespan should come from configuration.

src/SFA.DAS.EmployerAccounts/Queries/GetClientContent/GetClientContentRequestValidator.cs
```diff
@@ -0,0 +1,37 @@
...
+        {
+            var validationResult = new ValidationResult();
+
+            if (string.IsNullOrEmpty(item.ContentType))
+            {
+                validationResult.AddError(nameof(item.ContentType), "Type has not been supplied");
+            }
+            else
+            {
+                var isValidType = Enum.TryParse(item.ContentType, true, out ContentType type);
```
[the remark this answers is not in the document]
shomavg: To ensure that the conversion from string to Enum does not fail and that only valid content types is allowed.

src/SFA.DAS.EmployerAccounts.Web/Extensions/HtmlHelperExtensions.cs
shomavg: This is the eg: format ?type=banner&clientId=eas-acc_legacy

src/SFA.DAS.EmployerAccounts/Interfaces/ICacheStorageService.cs
shomavg: Im just following the pattern that was used in the other repos.

src/SFA.DAS.EmployerAccounts/Services/CacheStorageService.cs
shomavg: The story has mentioned that it is going to always be 1 hr.

src/SFA.DAS.EmployerAccounts.Web/Extensions/HtmlHelperExtensions.cs
cofaulco: I think we could store this in config. The clientId will remain unchanged for each calling application

src/SFA.DAS.EmployerAccounts/Interfaces/ICacheStorageService.cs
cofaulco: I don't mind this being changed to Get, Save and Delete.

src/SFA.DAS.EmployerAccounts/Queries/GetClientContent/GetClientContentRequestValidator.cs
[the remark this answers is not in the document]
cofaulco: is there any value in converting to an enum? The api just accepts the content type as a string

src/SFA.DAS.EmployerAccounts/Queries/GetClientContent/GetClientContentRequestHandler.cs
cofaulco: TryGet method as mentioned by @Stephek would make sense here

src/SFA.DAS.EmployerAccounts/Configuration/EmployerAccountsConfiguration.cs
```diff
@@ -49,5 +49,8 @@ public class EmployerAccountsConfiguration : ITopicMessagePublisherConfiguration
         public CommitmentsApiV2ClientConfiguration CommitmentsApi { get; set; }
         public RecruitClientApiConfiguration RecruitApi { get; set; }
         public long DefaultServiceTimeoutMilliseconds { get; set; }
+        public ContentClientApiConfiguration ContentClientApi { get; set; }
+        public string ApplicationId { get; set; }
+        public int DefaultCacheExpirationInHours { get; set; }
```
Stephek, line 54: Would minutes be better?

src/SFA.DAS.EmployerAccounts/Interfaces/IClientContentService.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System.Threading.Tasks;
+
+namespace SFA.DAS.EmployerAccounts.Interfaces
+{
+    public interface IClientContentService
+    {
+        Task<string> GetContent(string type, string clientId);
```
Stephek, line 7: Can we name this method Get rather than GetContent.  It's the client content service so we would expect it to return client content.

src/SFA.DAS.EmployerAccounts/Interfaces/ICacheStorageService.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System.Threading.Tasks;
+
+namespace SFA.DAS.EmployerAccounts.Interfaces
+{
+    public interface ICacheStorageService
+    {
+        //Task<T> TryGet<T>(string key);
```
Stephek, line 7: Delete commented out code.

src/SFA.DAS.EmployerAccounts/Interfaces/ICacheStorageService.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System.Threading.Tasks;
+
+namespace SFA.DAS.EmployerAccounts.Interfaces
+{
+    public interface ICacheStorageService
```
Stephek, line 5: Much cleaner interface.

src/SFA.DAS.EmployerAccounts/Interfaces/IClientContentService.cs
Stephek, line 7: Is clientId used?

src/SFA.DAS.EmployerAccounts/Interfaces/IClientContentApiClient.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System.Threading.Tasks;
+
+namespace SFA.DAS.EmployerAccounts.Interfaces
+{
+    public interface IClientContentApiClient
+    {
+        Task<string> GetContent(string type, string clientId);
```
Stephek, line 7: ClientId or ApplicationId?

src/SFA.DAS.EmployerAccounts/Services/ClientContentApiClient.cs
```diff
@@ -0,0 +1,44 @@
...
+        }
+
+        public Task<string> GetContent(string type, string clientId)
+        {
+            string banner =string.Empty;
+            switch (type)
+            {
+                case "banner":
+                    {
+                        if (clientId.Contains("_legacy"))
```
Stephek, line 21: Seems to be confusion between clientId and applicationId.  Maybe just my confusion?

src/SFA.DAS.EmployerAccounts/Interfaces/ICacheStorageService.cs
```diff
@@ -0,0 +1,11 @@
...
+    {
+        Task Save<T>(string key, T item, int expirationInHours);
```
Stephek, line 7: Please rename parameter to expirationInMinutes (or better pass in a Timespan)

PR #2088, opened 2020-05-28
CON-1871-add zendesk api push code.

shomavg:


src/SFA.DAS.EmployerAccounts.Web/App_Start/FilterConfig.cs
```diff
@@ -18,7 +19,8 @@ public static void RegisterGlobalFilters(GlobalFilterCollection filters)
             filters.AddUnitOfWorkFilter();
             filters.Add(new GoogleAnalyticsFilter());            
             filters.AddDasEmployerAccountsAuthorizationFilter();
-            filters.AddDasEmployerAccountsUnauthorizedAccessExceptionFilter();            
+            filters.AddDasEmployerAccountsUnauthorizedAccessExceptionFilter();
+            filters.Add(new AnalyticsFilter());
```
cofaulco, line 23: do we still need the GoogleAnalyticsFilter if we have "AnalyticsFilter"?

PR #2090, opened 2020-06-01
CON-1920-Added logs

VasanthaKasirajan3008:


src/SFA.DAS.EAS.Support.Web/Controllers/AccountController.cs
```diff
@@ -119,14 +145,20 @@ public async Task<ActionResult> Finance(string id)
...
         }
 
         [Route("account/levysubmissions/{id}/{payeSchemeId}")]
         public async Task<ActionResult> PayeSchemeLevySubmissions(string id, string payeSchemeId)
         {
+            _log.Info($"AccountController-PayeSchemeLevySubmissions : Getting PayeLevySubmissionsResponseCodes Response for id : {id}");
+
             var response = await _payeLevySubmissionsHandler.FindPayeSchemeLevySubmissions(id, payeSchemeId);
 
+            _log.Info($"AccountController-PayeSchemeLevySubmissions : PayeLevySubmissionsResponseCodes Response : {response}");
```
cofaulco, line 160: logging the response won't log any of the response data

VasanthaKasirajan3008: Removed the logs as we discussed.

PR #2100, opened 2020-06-12
CON-1795-update package for nservicebus

shomavg:


cofaulco:
no longer deemed necessary as the fix this is trying to implement is already in the current version of the code.

PR #2106, opened 2020-07-20
Migrate to das ci

CodeWolf86:


CodeWolf86:
For Reference, Regression passed in Test environment with changes. https://sfa-gov-uk.visualstudio.com/Digital%20Apprenticeship%20Service/_releaseProgress?releaseId=44183&_a=release-environment-extension&environmentId=212490&extensionId=ms.vss-test-web.test-result-in-release-environment-editor-tab

PR #2108, opened 2020-08-04
Con 1493 expire legal agreements

Howesy77:


CodeWolf86:
As an addendum to the above. You might want to wrap it in a transaction so it rolls back in case any issues crop up. Would be a pain to unpick even if it is unlikely to break.

cofaulco:
Branch was part of v4 agreement feature which was merged to master. Once updated with the lates master there are no changes in this branch. Closing PR and deleting the branch

PR #2122, opened 2020-09-03
Web config change to fix cookie samesite issue

Howesy77:


cofaulco (approved):
Raising tech debt to address this issue properly in code. But as it is a potential PROD blocker approving as is

PR #2133, opened 2020-10-05
CON-2349-fix non-owner issue

shomavg:


src/SFA.DAS.EmployerAccounts/Commands/SignEmployerAgreement/SignEmployerAgreementCommandHandler.cs
```diff
@@ -34,7 +38,9 @@ public class SignEmployerAgreementCommandHandler : IAsyncRequestHandler<SignEmpl
         private readonly IMediator _mediator;
         private readonly IEventPublisher _eventPublisher;
         private readonly ICommitmentService _commitmentService;
-        
+        private readonly Lazy<EmployerAccountsDbContext> _database;
```
Howesy77, line 41: Should probably use repo layer

Howesy77 (changes requested):

Howesy77 (approved):

PR #2137, opened 2020-10-27
CON-2531 access denied when view or transactor attempts to remove org

ben1stone:


src/SFA.DAS.EmployerAccounts.Web/Controllers/EmployerAgreementController.cs
```diff
@@ -222,6 +222,11 @@ public async Task<ActionResult> ConfirmRemoveOrganisation(string accountLegalEnt
         {
             var model = await _orchestrator.GetConfirmRemoveOrganisationViewModel(accountLegalEntityHashedId, hashedAccountId, OwinWrapper.GetClaimValue(ControllerConstants.UserRefClaimKeyName));
 
+            if (model.Status == HttpStatusCode.Unauthorized)
```
cofaulco, line 225: I think this should be handled by the base controller. I think the only change needed to be made is to pass `model` rather than `model.data` to the `View()`

cofaulco (changes requested):

src/SFA.DAS.EmployerAccounts.Web/Controllers/EmployerAgreementController.cs
```diff
@@ -222,7 +222,7 @@ public async Task<ActionResult> ConfirmRemoveOrganisation(string accountLegalEnt
         {
             var model = await _orchestrator.GetConfirmRemoveOrganisationViewModel(accountLegalEntityHashedId, hashedAccountId, OwinWrapper.GetClaimValue(ControllerConstants.UserRefClaimKeyName));
 
-            return View(model.Data.CanBeRemoved ? ControllerConstants.ConfirmRemoveOrganisationActionName : ControllerConstants.CannotRemoveOrganisationViewName, model.Data);
+            return View(model.Data != null && model.Data.CanBeRemoved ? ControllerConstants.ConfirmRemoveOrganisationActionName : ControllerConstants.CannotRemoveOrganisationViewName, model);
```
cofaulco, line 225: These 2 views would need to be changed to expect and OrchestrationResponse<> rather than the ViewModel from the  Data property.

cofaulco (changes requested):

cofaulco (approved):

PR #2142, opened 2020-12-07
Redirect to access denied page in not authorised

Howesy77:


src/SFA.DAS.EmployerAccounts.Web/Controllers/OrganisationController.cs
```diff
@@ -85,31 +86,26 @@ public async Task<ActionResult> OrganisationAddedNextStepsSearch(string organisa
 
             var response = await _orchestrator.CreateLegalEntity(request);
 
+            if (response.Status == HttpStatusCode.Unauthorized)
+            {
+                return View(ControllerConstants.AccessDeniedViewName);
```
cofaulco, line 91: If you just return `View(response)` then the BaseController should handle the response and redirect, also adding the accountId to the ViewBag

Howesy77: Done

cofaulco (approved):

PR #2157, opened 2021-01-21
EI-818-add-signed-agreement-version-to-api-model

Stephek:


MichaelYoung1981 (approved):

MichaelYoung1981 (changes requested):
The event we use already had the version on it.

PR #2177, opened 2021-04-30
[TM-12] Adds Your transfers page & wraps it in a feature toggle

DannyGB:
Adds the "Your transfers" page and wraps it in a feature toggle.
You will need the https://github.com/SkillsFundingAgency/das-employer-config/tree/TM-12-add-leavy-transfer-matching-feature branch config

### DIFF
--- a/src/SFA.DAS.EmployerAccounts.Web/Extensions/UrlHelperExtensions.cs
+++ b/src/SFA.DAS.EmployerAccounts.Web/Extensions/UrlHelperExtensions.cs
@@ -54,14 +54,6 @@ namespace SFA.DAS.EmployerAccounts.Web.Extensions
             return Action(baseUrl, $"{hashedAccountId}/{path}");
         }
 
-        public static string YourTransfersAction(this UrlHelper helper, string path = "")
-        {
-            var configuration = DependencyResolver.Current.GetService<EmployerAccountsConfiguration>();
-            var baseUrl = configuration.EmployerFinanceBaseUrl;
-            var hashedAccountId = helper.RequestContext.RouteData.Values[ControllerConstants.AccountHashedIdRouteKeyName];
-            return Action(baseUrl, $"transfers/{hashedAccountId}");
-        }
-
         public static string EmployerProjectionsAction(this UrlHelper helper, string path)
         {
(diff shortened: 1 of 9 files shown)


src/SFA.DAS.EmployerAccounts.Web/Views/EmployerTeam/Index.cshtml
```diff
@@ -102,6 +102,16 @@ else
                             <p>Apply for the payment, view applications and change your payment details.</p>
                         </div>
                     }
+
+                    @if (Model.Data.ApprenticeshipEmployerType!= ApprenticeshipEmployerType.Levy
```
chrisfoster76, line 106: Shouldn't this be `==` as we want to show the link to levy payers only?

src/SFA.DAS.EmployerAccounts.Web/Extensions/UrlHelperExtensions.cs
```diff
@@ -54,6 +54,14 @@ public static string EmployerIncentivesAction(this UrlHelper helper, string path
             return Action(baseUrl, $"{hashedAccountId}/{path}");
         }
 
+        public static string YourTransfersAction(this UrlHelper helper, string path = "")
```
chrisfoster76, line 57: Is this required? Couldn't we just make use of `EmployerFinanceAction` as the endpoint sits within Finance?

chrisfoster76 (changes requested):
Minor possible changes

src/SFA.DAS.EmployerAccounts.Web/Extensions/UrlHelperExtensions.cs
DannyGB: Yes, I'll re-use the existing `EmployerFinanceAction` action

src/SFA.DAS.EmployerAccounts.Web/Views/EmployerTeam/Index.cshtml
DannyGB: Yes it should

chrisfoster76 (approved):

PR #2181, opened 2021-05-18
[TM-13] Link off to LTM Sub-Site

mmiddleton3301:
Creates a link off to the new LTM sub-site. Only displays if the user is an Owner or Transactor. Includes new config.

Also see:

* https://github.com/SkillsFundingAgency/das-employer-config/pull/871
* https://github.com/SkillsFundingAgency/das-levy-transfer-matching-web/pull/14

### DIFF
--- a/src/SFA.DAS.EmployerFinance.Web/Controllers/TransfersController.cs
+++ b/src/SFA.DAS.EmployerFinance.Web/Controllers/TransfersController.cs
@@ -8,11 +8,11 @@ namespace SFA.DAS.EmployerFinance.Web.Controllers
     [DasAuthorize("EmployerFeature.TransfersMatching")]
     [RoutePrefix("accounts/{HashedAccountId}")] public class TransfersController : Controller
     {
-        private readonly TransfersOrcestrator _transfersOrcestrator;
+        private readonly TransfersOrchestrator _transfersOrcestrator;
 
-        public TransfersController(TransfersOrcestrator transfersOrcestrator)
+        public TransfersController(TransfersOrchestrator transfersOrchestrator)
         {
-            _transfersOrcestrator = transfersOrcestrator;
+            _transfersOrcestrator = transfersOrchestrator;
         }
 
         [HttpGet]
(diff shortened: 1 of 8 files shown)


src/SFA.DAS.EmployerFinance.Web/Controllers/TransfersController.cs
```diff
@@ -1,16 +1,27 @@
...
+using System.Web.Mvc;
 using SFA.DAS.Authorization.Mvc.Attributes;
+using SFA.DAS.EmployerFinance.Web.Orchestrators;
 
 namespace SFA.DAS.EmployerFinance.Web.Controllers
 {
     [DasAuthorize("EmployerFeature.TransfersMatching")]
     [RoutePrefix("accounts/{HashedAccountId}")] public class TransfersController : Controller
     {
+        private readonly TransfersOrcestrator _transfersOrcestrator;
```
chrisfoster76, line 11: Typo: missing "h"

src/SFA.DAS.EmployerFinance.Web/Views/Transfers/Index.cshtml
```diff
@@ -9,6 +10,13 @@
 
 <h1 class="heading-xlarge">Your Transfers</h1>
 
+@if (Model.Data.RenderCreateTransfersPledgeButton)
+{
+<div class="govuk-button-group">
+    <a class="govuk-button" data-module="govuk-button" href="@Url.LevyMatchingTransfersAction("pledges")">Create a transfers pledge</a>
```
chrisfoster76, line 16: Can we add an id attribute to this button, mostly so that automated tests can find it easily.

chrisfoster76 (approved):
Approved with two trivial notes

src/SFA.DAS.EmployerFinance.Web/Controllers/TransfersController.cs
mmiddleton3301: Good spot.

PR #2190, opened 2021-07-08
Con 2327 manage identity content api client

VasanthaKasirajan3008:


src/SFA.DAS.EmployerAccounts.UnitTests/App.config
```diff
@@ -0,0 +1,6 @@
+﻿<?xml version="1.0" encoding="utf-8" ?>
+<configuration>
+  <appSettings>
+    <add key="EnvironmentName" value="LOCAL"/>
```
cofaulco, line 4: could set this in whichever tests require it rather than adding the config: `ConfigurationManager.AppSettings["EnvironmentName"] = "LOCAL";`

This way could test behaviour when = "NOT_LOCAL"

cofaulco (commented):
1 comment

src/SFA.DAS.EmployerAccounts.UnitTests/App.config
VasanthaKasirajan3008: Done 👍

cofaulco (approved):

PR #2198, opened 2021-08-25
TM-129 Adding non-levy content for transfers section and changing Transfers …

ConorTill:
…page title

### DIFF
--- a/src/SFA.DAS.EmployerAccounts.Web/Views/EmployerTeam/Index.cshtml
+++ b/src/SFA.DAS.EmployerAccounts.Web/Views/EmployerTeam/Index.cshtml
@@ -121,14 +121,6 @@ else
                         }
                     }
 
-                    @if (Model.Data.ApprenticeshipEmployerType == ApprenticeshipEmployerType.Levy
-                         && Html.IsAuthorized("EmployerFeature.TransfersMatching"))
-                    {
-                        <div class="card">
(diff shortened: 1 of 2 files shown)


src/SFA.DAS.EmployerAccounts.Web/Views/EmployerTeam/Index.cshtml
```diff
@@ -103,15 +103,25 @@ else
...
 
-                    @if (Model.Data.ApprenticeshipEmployerType == ApprenticeshipEmployerType.Levy 
-                         && Html.IsAuthorized("EmployerFeature.TransfersMatching"))
+                    @if (Html.IsAuthorized("EmployerFeature.TransfersMatching"))
                     {
-                        <div class="card">
-                            <h3 class="heading-medium"><a href="@Url.EmployerFinanceAction("transfers")">Your transfers</a></h3>
-                            <p>View your 25% transfers balance and manage your transfers, including finding a business to transfer your levy to.</p>
-                        </div>
+                        if (Model.Data.ApprenticeshipEmployerType == ApprenticeshipEmployerType.Levy)
```
sscaife, line 108: The url is the same is that correct? other than that looks good

sscaife (approved):

src/SFA.DAS.EmployerAccounts.Web/Views/EmployerTeam/Index.cshtml
ConorTill: Yep this url should be the same

PR #2203, opened 2021-09-02
Tm 133 hide the option to create a transfers pledge from levy receivers and non levy employers

sscaife:


src/SFA.DAS.EmployerFinance/DependencyResolution/ManageApprenticeshipsOuterApiRegistry.cs
```diff
@@ -12,6 +13,7 @@ public ManageApprenticeshipsOuterApiRegistry ()
         {
             For<ManageApprenticeshipsOuterApiConfiguration>().Use(c => c.GetInstance<EmployerFinanceConfiguration>().ManageApprenticeshipsOuterApiConfiguration).Singleton();
             For<IApiClient>().Use<ApiClient>().Ctor<HttpClient>().Is(new HttpClient()).Singleton();
+            For<ICohortsService>().Use<CohortsService>().Ctor<HttpClient>().Is(new HttpClient()).Ctor<string>().Is(c => c.GetInstance<EmployerFinanceConfiguration>().EmployerCommitmentsBaseUrl).Singleton();
```
ConorTill, line 16: If we can use the same pattern as LevyTransferMatchingService maybe it can be injected the same way?

src/SFA.DAS.EmployerFinance.Web/Views/Transfers/Index.cshtml
```diff
@@ -8,23 +8,26 @@
...
-        <div class="panel panel-border-wide">
-            You have <strong>@Model.Data.PledgesCount</strong> transfer pledges.
-        </div>
-        <div class="">
-            @if (Model.Data.RenderCreateTransfersPledgeButton)
-            {
-                <a class="button" href="@Url.LevyMatchingTransfersAction("pledges/create/inform")" id="CreateTransfersPledgeButton">Create a transfers pledge</a>
-            }
-            <a href="@Url.LevyMatchingTransfersAction("pledges")" class="button button-secondary">View my transfer pledges and applications</a>
+@if (Model.Data.CanViewTransfersSection)
```
ConorTill, line 11: Looks like this is hiding the content of the entire page. From my understanding of the acceptance criteria we should just be hiding the 'My pledges' section. If this is the case we should probably rename the property to CanViewPledgesSection too.

ConorTill (changes requested):

src/SFA.DAS.EmployerFinance.Web/Orchestrators/TransfersOrchestrator.cs
```diff
@@ -21,16 +25,40 @@ protected TransfersOrchestrator()
...
+            _cohortsService = cohortsService;
         }
 
         public async Task<OrchestratorResponse<TransfersIndexViewModel>> Index(string hashedAccountId)
         {
-            bool renderCreateTransfersPledgeButton = await _authorizationService.IsAuthorizedAsync(EmployerUserRole.OwnerOrTransactor);
+            var accountDetail = await _accountApiClient.GetAccount(hashedAccountId).ConfigureAwait(false);
+            var cohortsCount = await _cohortsService.GetCohortsCount(accountDetail.AccountId).ConfigureAwait(false);
+
+            if (string.Compare(accountDetail.ApprenticeshipEmployerType, "Levy", StringComparison.OrdinalIgnoreCase) != 0 || cohortsCount > 0)
```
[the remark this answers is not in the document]
sscaife: that's good to know, cheers

src/SFA.DAS.EmployerFinance.Web/Views/Transfers/Index.cshtml
sscaife: moved the h1 and renamed the property as we discussed

chrisfoster76 (approved):

PR #2205, opened 2021-09-08
CON-3854 - Organisation can be deleted

Howesy77:


src/SFA.DAS.EmployerAccounts/Queries/GetAccountLegalEntityRemove/GetAccountLegalEntityRemoveQueryHandler.cs
```diff
@@ -84,6 +84,7 @@ private async Task<bool> SetRemovedStatusBasedOnCommitments(long accountId, Acco
 
             return commitmentConnectedToEntity == null || (commitmentConnectedToEntity.ActiveCount +
                                                            commitmentConnectedToEntity.PendingApprovalCount +
+                                                           commitmentConnectedToEntity.CompletedCount +
```
cofaulco, line 87: I think you need to use WithdrawnCount to include Stopped apprenticeships. Withdrawn = Stopped. Although completed is probably also a valid check.

cofaulco (changes requested):
1 change

src/SFA.DAS.EmployerAccounts/Queries/GetAccountLegalEntityRemove/GetAccountLegalEntityRemoveQueryHandler.cs
Howesy77: Fixed

cofaulco (approved):

PR #2208, opened 2021-09-30
Adds "internal" api endpoint to get transfer connections

chrisfoster76:


src/SFA.DAS.EmployerAccounts/Queries/GetTransferConnections/GetTransferConnectionsQuery.cs
```diff
@@ -4,6 +4,7 @@ namespace SFA.DAS.EmployerAccounts.Queries.GetTransferConnections
 {
     public class GetTransferConnectionsQuery : IAsyncRequest<GetTransferConnectionsResponse>
     {
+        public long? AccountId { get; set; }
```
cofaulco, line 7: still here

cofaulco (approved):

PR #2211, opened 2021-10-01
CON-3951 - Revamped legal agreements page

Howesy77:


src/SFA.DAS.EmployerAccounts.Web/Views/EmployerAgreement/ViewAllAgreements.cshtml
```diff
@@ -8,109 +9,35 @@
...
+            <div class="govuk-accordion" data-module="govuk-accordion" id="accordion-default">
+                @foreach (var agreement in Model.Data.Where(agreement => agreement.SignedDate != null))
                 {
-                    tabsCount++;
-                    <li class="govuk-tabs__list-item@(tabsCount == 1 ? "" : " govuk-tabs__list-item--selected")">
-                        <a class="govuk-tabs__tab" href="@item.GetAgreementTabListId">
-                            @item.Template.PublishedInfo
-                        </a>
-                    </li>
+                    @Html.Partial("_AcceptedAgreement", agreement, new ViewDataDictionary{{"Inset", agreement.Template.InsetText(Model.Data) } })
```
cofaulco, line 33: Is it worth using some sort of constant rather than hardcoding "Inset"

src/SFA.DAS.EmployerAccounts.Web/Controllers/EmployerAgreementController.cs
```diff
@@ -292,6 +292,8 @@ public async Task<ActionResult> WhenDoYouWantToView(int? choice, string agreemen
         public async Task<ActionResult> ViewAllAgreements(string hashedAccountId, string accountLegalEntityHashedId)
         {
             var model = await _orchestrator.GetOrganisationAgreements(accountLegalEntityHashedId);
+            ViewBag.AgreementId = accountLegalEntityHashedId;
```
cofaulco, line 295: would it be better to have the "AgreementId" on the model rather than using viewbag?

cofaulco (commented):
a couple of minor comments

cofaulco (approved):

PR #2215, opened 2021-10-07
CON-4039 - Sign out link in Finance takes to page not found

Howesy77:


src/SFA.DAS.EmployerFinance.Web/Extensions/HtmlHelperExtensions.cs
```diff
@@ -65,7 +65,7 @@ public static IHeaderViewModel GetHeaderViewModel(this HtmlHelper html, bool use
                 AuthenticationAuthorityUrl = configuration.Identity.BaseAddress,
                 ClientId = configuration.Identity.ClientId,
                 EmployerRecruitBaseUrl = configuration.EmployerRecruitBaseUrl,
-                SignOutUrl = new Uri($"{configuration.EmployerFinanceBaseUrl}/services/signOut/"),
+                SignOutUrl = new Uri($"{configuration.EmployerFinanceBaseUrl}/service/signOut/"),
```
cofaulco, line 68: should we check for `.EndsWith("/")` ?

PR #2229, opened 2021-12-23
CON-4311-EIUpdatesAndContentChanges

hkf-tech:


src/SFA.DAS.EmployerAccounts/Services/ContentApiClient.cs
```diff
@@ -29,10 +29,15 @@ public async Task<string> Get(string type, string applicationId)
             await AddAuthenticationHeader();
 
             string banner = string.Empty;
+            var uri = string.Empty;
             switch (type)
```
cofaulco, line 33: is this switch necessary? Always doing the same thing.

cofaulco (commented):
1 comment

src/SFA.DAS.EmployerAccounts/Services/ContentApiClient.cs
hkf-tech: I was thinking the same, was just following the pattern. Can take it out

cofaulco (approved):

PR #2252, opened 2022-02-08
TM-263 Remove IsTransferReceiver and CanViewPledgeSection from manage transf…

ConorTill:
…ers index


src/SFA.DAS.EmployerFinance.Web/Orchestrators/TransfersOrchestrator.cs
```diff
@@ -57,7 +57,6 @@ public async Task<OrchestratorResponse<IndexViewModel>> GetIndexViewModel(string
             {
                 Data = new IndexViewModel
                 {
-                    CanViewPledgesSection = !(indexTask.Result.IsTransferReceiver || employerType == ApprenticeshipEmployerType.NonLevy),
```
jawwadbaig, line 60: What will happen if the type is Apprenticeship Employer NonLevey?

PR #2255, opened 2022-02-09
Tm 59 transfer funds remaining

LACrossley:
Added RemainingTransferFunds to transfers view, view model, and TransfersOrchestrator.


SHESFA (approved):

src/SFA.DAS.EmployerFinance/Infrastructure/OuterApiResponses/Transfers/GetIndexResponse.cs
```diff
@@ -4,5 +4,6 @@ public class GetIndexResponse
     {
         public int PledgesCount { get; set; }
         public int ApplicationsCount { get; set; }
+        public int ActivePledgesTotalAmount { get; set; }
```
jawwadbaig: Sorry, does this field needs to be decimal as well or this one is okay? -->> public int ActivePledgesTotalAmount

jawwadbaig (approved):

PR #2256, opened 2022-02-10
commtented use installer as this was causing to reset the size of the…

hkf-tech:
… queue to 1GB


hardik-desai-kcl (approved):

dashton82 (approved):
Need to look into a long term fix, but this will stop the queue size from being reset on deployment

MichaelYoung1981 (approved):

PR #2259, opened 2022-02-18
CON-4487 - Allow organisations registered via AORN to update their details

Howesy77:


src/SFA.DAS.EmployerAccounts/Interfaces/IPensionRegulatorService.cs
```diff
@@ -6,8 +6,8 @@ namespace SFA.DAS.EmployerAccounts.Interfaces
 {
     public interface IPensionRegulatorService
     {
-        Task<IEnumerable<Organisation>> GetOrgansiationsByPayeRef(string payeRef);
-
-        Task<IEnumerable<Organisation>> GetOrgansiationsByAorn(string aorn, string payeRef);
+        Task<Organisation> GetOrganisationById(string payeRef);
```
cofaulco, line 9: should the param name here be organisationId or something similar?

Howesy77: Yep!

cofaulco (approved):

cofaulco (changes requested):
should be some new unit tests for `GetOrganisationByIdQueryHandler.cs`

cofaulco (approved):

PR #2272, opened 2022-04-01
CON-4625 Support Console Add Levy status

kola-samuel:


src/SFA.DAS.EAS.Support.Web/Views/Account/SubHeader.cshtml
```diff
@@ -5,12 +5,20 @@
...
+            </a>
+        </h5>
+
         <h1 class="heading-large">
             @(Model?.DasAccountName)
             @if (!string.IsNullOrEmpty(Model?.PublicHashedAccountId))
             {
                 <span class="heading-secondary"> Account ID @(Model?.PublicHashedAccountId), created @Model.DateRegistered.ToString("dd/MM/yyyy")</span>
             }
+            <span class="heading-secondary"> Levy Status  @Model.ApprenticeshipEmployerType</span>
```
cofaulco, line 21: I think the LEVY or NON-LEVY value is supposed to be in a tag:
https://design-system.service.gov.uk/components/tag/

as per the design on the ticket.

Support console uses the old gov-uk stylesheets however so will need to find the correct style to apply.

cofaulco (changes requested):
1 change required

PR #2288, opened 2022-05-03
Tm 44 transfer funds committed future financial years web

jawwadbaig:
Tm 44 transfer funds committed future financial years web


ConorTill (approved):

src/SFA.DAS.EmployerFinance/Extensions/DateTimeExtensions.cs
```diff
@@ -23,5 +19,31 @@ public static string ToFinancialYearString(this DateTime dateTime)
                 return $"{dateTime.Year}/{dateTime.AddYears(1).ToString("yy")}";
             }
         }
+
+        public static string ToNextFinancialYearString(this DateTime dateTime)
```
chrisfoster76, line 23: There's a lot of similar code here - wouldn't it be better to just have a single extension method to give a financial year display label given a date, and just call that multiple times with different dates? The original spec was for a 5 year projection - we wouldn't want to just keep adding more and more extension methods to give us f/y labels for all those future years.

jawwadbaig: yes sure. I have removed the extra extension methods and kept just one and passed the date parameter.

chrisfoster76: This isn't what I meant, sorry:
```
public static string ToFinancialYearString(this DateTime dateTime, DateTime date)
```

It's impossible to tell what this method does just from the signature.

What I meant was that we should just have an extension method off a DateTime that gives the FinancialYear display string for financial year into which that date falls. _Outside of this_, in the calling code, we can call it multiple times using a DateTime instance, then that same value +1 year, then the same value +2 years, etc.

Local variables should be named lowercase and declararion with var is the established pattern. Also you can express the format string using `:` instead of the `ToString()` method.

chrisfoster76 (changes requested):

PR #2294, opened 2022-05-19
update api for MI auth and update client package to use either client…

cofaulco:
… secret or MI to obtain bearer token.

### DIFF
--- a/azure/finance.template.json
+++ b/azure/finance.template.json
@@ -94,6 +94,12 @@
         },
         "tags": {
             "type": "object"
+        },
+        "idaTenant": {
+            "type": "string"
+        },
+        "financeApiIdentifierUri": {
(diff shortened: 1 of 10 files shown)


ChrisJWoodcock (changes requested):
It looks like this isn't configured to be have audiences currently, so we should be doing this in a consistent style with other apps e.g. typically this is done using an AzureActiveDirectoryConfiguration section in the config file. Since there are not any clients (?) of the Finance API currently we could take the opportunity to drop the client configuration from employer config, not setting up the dependency in the client library and reduce the configuration to only the MI (ApiBaseUrl and IdentifierUrl) parts from the start and not bother checking for supplied client credentials.

ChrisJWoodcock (approved):

PR #2301, opened 2022-05-30
Mac 16 recruit api mi account

VasanthaKasirajan3008:


src/SFA.DAS.EmployerAccounts.UnitTests/Services/Recruit/WhenIGetVacancies.cs
```diff
@@ -36,10 +36,10 @@ public void Arrange()
         {
             _hashedAccountId = Guid.NewGuid().ToString();
             _apiBaseUrl = $"http://{Guid.NewGuid().ToString()}";
-            _clientId = Guid.NewGuid().ToString();
-            _clientSecret = Guid.NewGuid().ToString();
+            _clientId = string.Empty;
```
cofaulco, line 39: could change the HttpServiceFactory to have optional params for the non-MI stuff now, which means we could delete all these empty variables as they are no longer required.

cofaulco (approved):

PR #2339, opened 2022-06-29
MAC-64_Store_user_from_claim_details_during_authentication

ChrisJWoodcock:


src/SFA.DAS.EmployerFinance.AcceptanceTests/Steps/Hooks.cs
```diff
@@ -1,129 +1,128 @@
-﻿using System;
-using System.Data.Common;
-using System.Diagnostics;
```
cofaulco, line 3: could we revert so system usings appear at the top of the list please?

cofaulco (changes requested):

cofaulco (approved):

cofaulco (changes requested):
couple of bits

cofaulco (approved):

cofaulco (approved):

PR #2389, opened 2022-09-07
FAI-605 - Remove SFA.DAS.NServiceBus dependency from messages

dashton82:
Update classes not to inherit from Event


ChrisJWoodcock (approved):
Assuming that the event has identical fields as previously.

cofaulco (approved):

PR #2409, opened 2022-12-02
Update to include hashing registry for finance jobs

cofaulco:


ChrisJWoodcock (approved):
Approved on the understanding that a subsequent release will revert the change to RunOnStartup setting it back to false to prevent it running after every release

PR #2411, opened 2022-12-05
duplicate batching logic used within the ExpireFunds job for draft ex…

cofaulco:
…piry handler


src/SFA.DAS.EmployerFinance.MessageHandlers/CommandHandlers/DraftExpireFundsCommandHandler.cs
```diff
@@ -21,19 +21,28 @@ public async Task Handle(DraftExpireFundsCommand message, IMessageHandlerContext
...
-
-                sendOptions.RequireImmediateDispatch();
                 sendOptions.RouteToThisEndpoint();
 
-                return context.Send(c, sendOptions);
-            });
+                messageTasks.Add(context.Send(new DraftExpireAccountFundsCommand { AccountId = account.Id, DateTo = message.DateTo }, sendOptions));
+                sendCounter++;
+
+                if (sendCounter % 1000 == 0)
```
ChrisJWoodcock, line 36: Using messageTasks.Count() == 1000 would be simpler

ChrisJWoodcock (dismissed):

dashton82 (approved):

PR #2447, opened 2023-04-24
Map 231 estimated transfer balance

ratheeshri1984:


src/SFA.DAS.EmployerFinance.Web/Views/Transfers/Index.cshtml
```diff
@@ -18,26 +18,36 @@
...
     {
-        <div class="column-two-thirds">
-            <h2 class="heading-medium">My transfer finances</h2>
-            <p>This is a financial snapshot of what is available.</p>
+        <div class="column-one-half">
+            <h2 class="heading-medium">Transfer allowance information.</h2>
             <div class="finance-highlight">
                 <span class="bold-small">
-                    Transfer allowance for @Model.Data.FinancialYearString
+                    Travel allowance as at 6 April @Model.Data.FinancialYearString
```
chrisfoster76, line 25: This might be a typo :)

src/SFA.DAS.EmployerFinance.Web/Orchestrators/TransfersOrchestrator.cs
```diff
@@ -41,10 +41,11 @@ public async Task<OrchestratorResponse<IndexViewModel>> GetIndexViewModel(string
             var accountId = _hashingService.DecodeValue(hashedAccountId);
             var indexTask = _transfersService.GetCounts(accountId);
             var accountDetail = _accountApiClient.GetAccount(hashedAccountId);
+            var financialBreakdownTask = _transfersService.GetFinancialBreakdown(accountId);
```
chrisfoster76, line 44: As appealing as it might be, you shouldn't make this request from here. The BFF pattern we're using means that each page in the UI should call one endpoint in the outer api (which in turn should only be referenced by one endpoint in the UI - a 1:1 ratio). This means you should add an endpoint to the outer api for this page if it doesn't already exist, and respond with all the information you need in its response.

chrisfoster76: It looks as though an api endpoint for it might already exist:

```
_transfersService.GetCounts(accountId)
```

Ridiculous name though.

https://github.com/SkillsFundingAgency/das-apim-endpoints/blob/master/src/EmployerFinance/SFA.DAS.EmployerFinance.Api/Controllers/TransfersController.cs#L27

chrisfoster76 (changes requested):

src/SFA.DAS.EmployerFinance.Web/Orchestrators/TransfersOrchestrator.cs
```diff
@@ -41,10 +41,11 @@ public async Task<OrchestratorResponse<IndexViewModel>> GetIndexViewModel(string
...
 
             var renderCreateTransfersPledgeButtonTask = _authorizationService.IsAuthorizedAsync(EmployerUserRole.OwnerOrTransactor);            
 
-            await Task.WhenAll(indexTask, renderCreateTransfersPledgeButtonTask, accountDetail);
+            await Task.WhenAll(indexTask, renderCreateTransfersPledgeButtonTask, accountDetail, financialBreakdownTask);
```
chrisfoster76, line 48: I may have misread this, but I thought that the indexTask was a call to an outer api endpoint, in which case the data we get from financialBreakdownTask ought to really come from there. There should be a 1:1 ratio between UI endpoints and outer api endpoints.

